PR #1, opened 2017-04-20
Support fetching staging slot URL in end-to-end test project

joelverhagen:
This mimics the script used in functional tests to allow running end-to-end tests against the gallery staging slot URL.

The parameters values live in Octopus and are passed to a VSTS build.

### DIFF
--- a/src/NuGet.Services.EndToEnd/BasicTests.cs
+++ b/src/NuGet.Services.EndToEnd/BasicTests.cs
@@ -8,7 +8,7 @@ using Xunit;
 
 namespace NuGet.Services.EndToEnd
 {
-    public class BasicTests
+    public class BasicTests : IClassFixture<TrustedHttpsCertificatesFixture>
     {
         [Fact]
         public async Task NuGetDotOrgIsReachableOverHttps()
@@ -20,4 +20,5 @@ namespace NuGet.Services.EndToEnd
             }
         }
     }
+
 }
--- a/src/NuGet.Services.EndToEnd/BasicTests.cs
+++ b/src/NuGet.Services.EndToEnd/BasicTests.cs
@@ -14,7 +14,7 @@ namespace NuGet.Services.EndToEnd
         public async Task NuGetDotOrgIsReachableOverHttps()
         {
             using (var httpClient = new HttpClient())
-            using (var response = await httpClient.GetAsync("https://www.nuget.org/"))
+            using (var response = await httpClient.GetAsync(EnvironmentSettings.GalleryBaseUrl))
             {
                 Assert.Equal(HttpStatusCode.OK, response.StatusCode);
             }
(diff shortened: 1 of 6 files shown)


src/NuGet.Services.EndToEnd/EnvironmentSettings.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+
+namespace NuGet.Services.EndToEnd
+{
+    public static class EnvironmentSettings
```
skofman1: if there is any code duplicated from FunctionalTests, please add a comment on the class, so we will know what needs to be consolidated

src/NuGet.Services.EndToEnd/BasicTests.cs
```diff
@@ -8,16 +8,17 @@
 
 namespace NuGet.Services.EndToEnd
 {
-    public class BasicTests
+    public class BasicTests : IClassFixture<TrustedHttpsCertificatesFixture>
```
skofman1: This class needs a less general name. When we add more tests, we will need clearity on where to put them..

src/NuGet.Services.EndToEnd/EnvironmentSettings.cs
```diff
@@ -0,0 +1,52 @@
...
+        };
+
+        public static string GalleryBaseUrl => GetEnvironmentVariable("GalleryBaseUrl", required: true);
+
+        public static string[] TrustedHttpsCertificates
+        {
+            get
+            {
+                var value = GetEnvironmentVariable("TrustedHttpsCertificates", required: false);
+                if (value == null)
```
scottbommarito:
```
var value = GetEnvironmentVariable("TrustedHttpsCertificates", required: false) ?? new string[0];
```

scottbommarito: Also I would probably name this variable something else--naming it `value` like a setter makes this slightly confusing.

src/NuGet.Services.EndToEnd/BasicTests.cs
joelverhagen: This is essentially a place-holder until we have real tests.

src/NuGet.Services.EndToEnd/EnvironmentSettings.cs
joelverhagen: Variable name -- sure.

Cannot coalesce `string` to `string[]`.

src/NuGet.Services.EndToEnd/EnvironmentSettings.cs
joelverhagen: Done.

src/NuGet.Services.EndToEnd/EnvironmentSettings.cs
joelverhagen: Done.

PR #4, opened 2017-04-27
Aggressively push end-to-end test packages to minimize test run time

joelverhagen:
If end-to-end tests need N packages total and the first test is running, push all N packages instead of just the one that the test needs. This means they will all start flowing through V3 at the same time and subsequent tests won't have to wait long for the package to be available.

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
+++ b/src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
@@ -25,17 +25,21 @@ namespace NuGet.Services.EndToEnd.Support
         private readonly SemaphoreSlim _pushLock;
         private readonly object _packagesLock = new object();
         private readonly IDictionary<PackageType, Package> _packages;
-        private readonly Clients _clients;
+        private readonly IGalleryClient _galleryClient;
         private readonly TestSettings _testSettings;
 
-        public PushedPackagesFixture()
+        public PushedPackagesFixture() : this(Clients.Initialize().Gallery, TestSettings.CreateFromEnvironment())
         {
-            _clients = Clients.Initialize();
-            _testSettings = TestSettings.CreateFromEnvironment();
+        }
+
+        public PushedPackagesFixture(IGalleryClient galleryClient, TestSettings testSettings)
+        {
+            _galleryClient = galleryClient;
+            _testSettings = testSettings;
             _pushLock = new SemaphoreSlim(initialCount: 1);
             _packages = new Dictionary<PackageType, Package>();
         }
-        
+
         public async Task<Package> PushAsync(PackageType requestedPackageType, ITestOutputHelper logger)
         {
             var pushedPackage = GetPushedPackage(requestedPackageType, logger);
@@ -119,7 +123,7 @@ namespace NuGet.Services.EndToEnd.Support
             {
                 using (var nupkgStream = new MemoryStream(package.NupkgBytes.ToArray()))
                 {
-                    await _clients.Gallery.PushAsync(nupkgStream);
+                    await _galleryClient.PushAsync(nupkgStream);
                 }
                 logger.WriteLine($"Package {package} has been pushed.");
             }
--- a/src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
+++ b/src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
@@ -11,74 +11,167 @@ using Xunit.Abstractions;
 
 namespace NuGet.Services.EndToEnd.Support
 {
+    /// <summary>
+    /// A test fixture for sharing pushed packages between tests. Since packages take quite a while (minutes) to make
+    /// their way through the V3 pipeline, packages should be pushed all at once and shared between tests, if possible.
+    /// </summary>
     public class PushedPackagesFixture : IDisposable
     {
-        private readonly object _packagesLock;
-        private readonly IDictionary<PackageType, PushedPackage> _packages;
+        private static readonly HashSet<PackageType> SemVer2PackageTypes = new HashSet<PackageType>
+        {
+            PackageType.SemVer2Prerelease,
+        };
+
+        private readonly SemaphoreSlim _pushLock;
+        private readonly object _packagesLock = new object();
+        private readonly IDictionary<PackageType, Package> _packages;
         private readonly Clients _clients;
+        private readonly TestSettings _testSettings;
 
         public PushedPackagesFixture()
         {
             _clients = Clients.Initialize();
-            _packagesLock = new object();
-            _packages = new Dictionary<PackageType, PushedPackage>();
+            _testSettings = TestSettings.CreateFromEnvironment();
+            _pushLock = new SemaphoreSlim(initialCount: 1);
+            _packages = new Dictionary<PackageType, Package>();
         }
         
-        public async Task<Package> PushAsync(PackageType packageType, ITestOutputHelper logger)
+        public async Task<Package> PushAsync(PackageType requestedPackageType, ITestOutputHelper logger)
         {
-            PushedPackage pushedPackage;
-            lock (_packagesLock)
+            var pushedPackage = GetPushedPackage(requestedPackageType, logger);
+            if (pushedPackage != null)
             {
-                if (!_packages.TryGetValue(packageType, out pushedPackage))
-                {
-                    var package = CreatePackage(packageType);
-                    pushedPackage = new PushedPackage(package);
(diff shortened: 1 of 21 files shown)


src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -11,108 +11,178 @@
...
-                if (pushedPackage.Pushed)
+                var packageTypes = GetPackageTypes(requestedPackageType);
+
+                // Push all of the package types that have not been pushed yet.
+                var pushTasks = new Dictionary<PackageType, Task<Package>>();
+                foreach (var packageType in packageTypes)
                 {
-                    logger.WriteLine($"Package {pushedPackage} has already been pushed.");
-                    return pushedPackage.Package;
+                    if (!_packages.TryGetValue(packageType, out Package package))
```
scottbommarito: Not using `_packagesLock` here? I see `GetPushedPackage` uses it on read.

Can we just use `GetPushedPackage` and check if it returns null?

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -11,108 +11,178 @@
...
+            _pushLock = new SemaphoreSlim(initialCount: 1);
+            _packages = new Dictionary<PackageType, Package>();
         }
         
-        public async Task<Package> PushAsync(PackageType packageType, ITestOutputHelper logger)
+        public async Task<Package> PushAsync(PackageType requestedPackageType, ITestOutputHelper logger)
         {
-            PushedPackage pushedPackage;
-            lock (_packagesLock)
+            var pushedPackage = GetPushedPackage(requestedPackageType, logger);
```
scottbommarito: could reduce this method to `return GetPushedPackage(requestedPackageType, logger) ?? await PushUnpushedPackagesAsync(requestedPackageType, logger);`

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -11,108 +11,178 @@
...
+                    {
+                        pushTasks.Add(packageType, UncachedPushAsync(requestedPackageType, packageType, logger));
+                    }
                 }
 
-                using (var packageStream = new MemoryStream(pushedPackage.Package.NupkgBytes.ToArray()))
+                await Task.WhenAll(pushTasks.Values);
+
+                // Add the pushed packages to the cache.
+                lock (_packagesLock)
```
scottbommarito: You could remove the code inside this `lock` if you added `_packages[requestedPackageType] = package` to `UncachedPushAsync`.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -11,108 +11,178 @@
...
-            if (pushedPackage.Pushed)
-            {
-                logger.WriteLine($"Package {pushedPackage} push has already been pushed.");
-                return pushedPackage.Package;
-            }
+            return await PushUnpushedPackagesAsync(requestedPackageType, logger);
+        }
 
-            var acquired = await pushedPackage.PushLock.WaitAsync(0);
+        private async Task<Package> PushUnpushedPackagesAsync(PackageType requestedPackageType, ITestOutputHelper logger)
```
scottbommarito: I think this function would read a lot easier if
- You had a function that checked if a single package was cached, and if not, pushed it and added it to the cache.
- This function called that function for all `PackageType`s and then returned `GetPushedPackage(requestedPackageType)`.

All the locks and caching could be handled inside the first method. This method just needs to make sure all packages are pushed and then check the cache.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -11,108 +11,178 @@
...
-                PushLock = new SemaphoreSlim(1);
-                Package = package;
-                Pushed = false;
-            }
+        }
 
-            public SemaphoreSlim PushLock { get; }
-            public Package Package { get; }
-            public bool Pushed { get; private set; }
+        private IEnumerable<PackageType> GetPackageTypes(PackageType requestedPackageType)
```
scottbommarito: There is currently no visibility here if `requestedPackageType` is one of the `SemVer2PackageTypes`. I think this should throw or at least log if that is the case.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
joelverhagen: That's a long line, I think this is easier to read.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -11,108 +11,178 @@
...
-                logger.WriteLine($"Package {pushedPackage} push has already been pushed.");
-                return pushedPackage.Package;
-            }
+            return await PushUnpushedPackagesAsync(requestedPackageType, logger);
+        }
 
-            var acquired = await pushedPackage.PushLock.WaitAsync(0);
+        private async Task<Package> PushUnpushedPackagesAsync(PackageType requestedPackageType, ITestOutputHelper logger)
+        {
+            var acquired = await _pushLock.WaitAsync(0);
```
joelverhagen: Never seen it before but I like it. It gives a more detailed test output.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -11,108 +11,178 @@
...
             try
             {
                 if (!acquired)
                 {
-                    logger.WriteLine($"Another test is already pushing {pushedPackage}. Waiting.");
-                    await pushedPackage.PushLock.WaitAsync();
-                    acquired = true;
```
joelverhagen: Yes, the initial `WaitAsync` will either set acquired to `true` (in which case this scope is not execute`) or to `false` in which case this scope is executed and `acquired` is still set to `true`.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -11,108 +11,178 @@
...
+                    logger.WriteLine("Another test is already pushing the packages. Waiting.");
+                    await _pushLock.WaitAsync();
```
joelverhagen: User's of `SemaphoreSlim` must assume that either `WaitAsync` throws and the lock is not actually acquired or it returns a correct `acquired` boolean (either `true` or `false`). If this assumption is made, then we are safe.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
joelverhagen: Yeah, good idea.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
joelverhagen: I wanted to be "correct". Enums can have integer values that do not correspond to a name.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
joelverhagen: The problem with this is that you will not be pushing in parallel.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
ryuyu: I'm a little confused. Where in this scope is acquired set to true? the line setting it to true explicitly appears to have been removed with your changeset?

joelverhagen: Discussed offline. Added the `acquired = true` back. Thanks guys.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
joelverhagen: Fixed.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
joelverhagen: Fixed.

joelverhagen:
Thought about this more. Added unit tests. I think this is a good precedent to set albeit not totally necessary in this case.

PR #5, opened 2017-04-27
Add Search Service E2E

ryuyu:
Adds a test for semverlevel filtering on search endpoints.

@joelverhagen i merged our changes together, but please give it the once over to make sure the merge went more or less correctly.

@xavierdecoster

### DIFF
--- a/src/NuGet.Services.EndToEnd/PushTests.cs
+++ b/src/NuGet.Services.EndToEnd/PushTests.cs
@@ -55,5 +55,30 @@ namespace NuGet.Services.EndToEnd
 
             await _clients.V3Search.WaitForPackageAsync(package.Id, package.Version, _logger);
         }
+
+        public async Task NewSemVer2PackageIsFilteredButAvailableInV3()
+        {
+            // Arrange
+            var package = await _pushedPackages.PushAsync(PackageType.SemVer2Prerelease, _logger);
+
+            // Wait for package to become available
+            await _clients.FlatContainer.WaitForPackageAsync(package.Id, package.Version, _logger);
+            await _clients.Registration.WaitForPackageAsync(package.Id, package.Version, _logger);
+            await _clients.V3Search.WaitForPackageAsync(package.Id, package.Version, _logger);
+
+            var shouldBeEmptyV3 = await _clients.V3Search.Query($"q={package.Id}");
+            var shouldBeEmptyV2 = await _clients.V3Search.SearchQuery($"q={package.Id}");
+            var shouldBeEmptyAutocomplete = await _clients.V3Search.Autocomplete($"q={package.Id}");
+            Assert.Equal(0, shouldBeEmptyV3.Data.Count);
+            Assert.Equal(0, shouldBeEmptyV2.Data.Count);
+            Assert.Equal(0, shouldBeEmptyAutocomplete.Data.Count);
+
+            var shouldNotBeEmptyV3 = await _clients.V3Search.Query($"q={package.Id}&semVerLevel=2.0.0");
+            var shouldNotBeEmptyV2 = await _clients.V3Search.SearchQuery($"q={package.Id}&semVerLevel=2.0.0");
+            var shouldNotBeEmptyAutocomplete = await _clients.V3Search.Autocomplete($"q={package.Id}&semVerLevel=2.0.0");
+            Assert.NotEqual(0, shouldNotBeEmptyV3.Data.Count);
+            Assert.NotEqual(0, shouldNotBeEmptyV2.Data.Count);
+            Assert.NotEqual(0, shouldNotBeEmptyAutocomplete.Data.Count);
+        }
     }
 }
--- a/src/NuGet.Services.EndToEnd/PushTests.cs
+++ b/src/NuGet.Services.EndToEnd/PushTests.cs
@@ -63,7 +63,9 @@ namespace NuGet.Services.EndToEnd
 
             // Wait for package to become available
             await _clients.FlatContainer.WaitForPackageAsync(package.Id, package.Version, _logger);
-            await _clients.Registration.WaitForPackageAsync(package.Id, package.Version, _logger);
+            // wait for package to be in both registrations
+            await _clients.Registration.WaitForPackageAsync(package.Id, package.Version, /*semver2*/ true, _logger);
+            await _clients.Registration.WaitForPackageAsync(package.Id, package.Version, /*semver2*/ false, _logger);
             await _clients.V3Search.WaitForPackageAsync(package.Id, package.Version, _logger);
 
             var shouldBeEmptyV3 = await _clients.V3Search.Query($"q={package.Id}");
--- a/src/NuGet.Services.EndToEnd/PushTests.cs
+++ b/src/NuGet.Services.EndToEnd/PushTests.cs
@@ -55,32 +55,28 @@ namespace NuGet.Services.EndToEnd
 
             await _clients.V3Search.WaitForPackageAsync(package.Id, package.Version, _logger);
         }
-
-        public async Task NewSemVer2PackageIsFilteredButAvailableInV3()
+        
+        [SemVer2Fact]
+        public async Task NewSemVer2PackageIsFiltered()
         {
             // Arrange
             var package = await _pushedPackages.PushAsync(PackageType.SemVer2Prerelease, _logger);
 
+            // Act
             // Wait for package to become available
-            await _clients.FlatContainer.WaitForPackageAsync(package.Id, package.Version, _logger);
-            // wait for package to be in both registrations
-            await _clients.Registration.WaitForPackageAsync(package.Id, package.Version, /*semver2*/ true, _logger);
-            await _clients.Registration.WaitForPackageAsync(package.Id, package.Version, /*semver2*/ false, _logger);
             await _clients.V3Search.WaitForPackageAsync(package.Id, package.Version, _logger);
 
             var shouldBeEmptyV3 = await _clients.V3Search.Query($"q={package.Id}");
-            var shouldBeEmptyV2 = await _clients.V3Search.SearchQuery($"q={package.Id}");
             var shouldBeEmptyAutocomplete = await _clients.V3Search.Autocomplete($"q={package.Id}");
-            Assert.Equal(0, shouldBeEmptyV3.Data.Count);
-            Assert.Equal(0, shouldBeEmptyV2.Data.Count);
-            Assert.Equal(0, shouldBeEmptyAutocomplete.Data.Count);
 
-            var shouldNotBeEmptyV3 = await _clients.V3Search.Query($"q={package.Id}&semVerLevel=2.0.0");
-            var shouldNotBeEmptyV2 = await _clients.V3Search.SearchQuery($"q={package.Id}&semVerLevel=2.0.0");
+            var shouldNotBeEmptyV3 = await _clients.V3Search.Query($"q=packageid:{package.Id}&semVerLevel=2.0.0&prerelease=true");
             var shouldNotBeEmptyAutocomplete = await _clients.V3Search.Autocomplete($"q={package.Id}&semVerLevel=2.0.0");
-            Assert.NotEqual(0, shouldNotBeEmptyV3.Data.Count);
-            Assert.NotEqual(0, shouldNotBeEmptyV2.Data.Count);
-            Assert.NotEqual(0, shouldNotBeEmptyAutocomplete.Data.Count);
+
+            // Assert
+            Assert.Equal(0, shouldBeEmptyV3.Data.Count);
+            Assert.Equal(0, shouldBeEmptyAutocomplete.Data.Count);
+            Assert.Equal(1, shouldNotBeEmptyV3.Data.Count);
+            Assert.Equal(1, shouldNotBeEmptyAutocomplete.Data.Count);
         }
     }
 }
--- a/src/NuGet.Services.EndToEnd/PushTests.cs
+++ b/src/NuGet.Services.EndToEnd/PushTests.cs
@@ -61,22 +61,26 @@ namespace NuGet.Services.EndToEnd
         {
             // Arrange
             var package = await _pushedPackages.PushAsync(PackageType.SemVer2Prerelease, _logger);
+            var searchBaseAddresses = await _clients.V3Index.GetSearchBaseUrls();
 
             // Act
             // Wait for package to become available
             await _clients.V3Search.WaitForPackageAsync(package.Id, package.Version, _logger);
 
-            var shouldBeEmptyV3 = await _clients.V3Search.Query($"q={package.Id}");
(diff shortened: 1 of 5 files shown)


src/NuGet.Services.EndToEnd/PushTests.cs
```diff
@@ -55,5 +55,32 @@ public PushTests(PushedPackagesFixture pushedPackages, ITestOutputHelper logger)
...
+
+        public async Task NewSemVer2PackageIsFilteredButAvailableInV3()
+        {
+            // Arrange
+            var package = await _pushedPackages.PushAsync(PackageType.SemVer2Prerelease, _logger);
+
+            // Wait for package to become available
+            await _clients.FlatContainer.WaitForPackageAsync(package.Id, package.Version, _logger);
+            // wait for package to be in both registrations
+            await _clients.Registration.WaitForPackageAsync(package.Id, package.Version, /*semver2*/ true, _logger);
```
joelverhagen: use real named parameters: `semVer2: true`

src/NuGet.Services.EndToEnd/PushTests.cs
```diff
@@ -55,5 +55,32 @@ public PushTests(PushedPackagesFixture pushedPackages, ITestOutputHelper logger)
...
+            await _clients.Registration.WaitForPackageAsync(package.Id, package.Version, /*semver2*/ false, _logger);
+            await _clients.V3Search.WaitForPackageAsync(package.Id, package.Version, _logger);
```
joelverhagen: You only need to wait for search right?

src/NuGet.Services.EndToEnd/PushTests.cs
```diff
@@ -55,5 +55,32 @@ public PushTests(PushedPackagesFixture pushedPackages, ITestOutputHelper logger)
 
             await _clients.V3Search.WaitForPackageAsync(package.Id, package.Version, _logger);
         }
+
+        public async Task NewSemVer2PackageIsFilteredButAvailableInV3()
```
joelverhagen: Just focus on `Filtered` or `AvailableInV3` not both. I think you just care about `Filtered`

src/NuGet.Services.EndToEnd/PushTests.cs
```diff
@@ -55,5 +55,32 @@ public PushTests(PushedPackagesFixture pushedPackages, ITestOutputHelper logger)
...
+            var package = await _pushedPackages.PushAsync(PackageType.SemVer2Prerelease, _logger);
+
+            // Wait for package to become available
+            await _clients.FlatContainer.WaitForPackageAsync(package.Id, package.Version, _logger);
+            // wait for package to be in both registrations
+            await _clients.Registration.WaitForPackageAsync(package.Id, package.Version, /*semver2*/ true, _logger);
+            await _clients.Registration.WaitForPackageAsync(package.Id, package.Version, /*semver2*/ false, _logger);
+            await _clients.V3Search.WaitForPackageAsync(package.Id, package.Version, _logger);
+
+            var shouldBeEmptyV3 = await _clients.V3Search.Query($"q={package.Id}");
```
joelverhagen: Please use Arrange/Act/Assert

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -0,0 +1,253 @@
...
+        {
+            public List<V2SearchPackage> Data { get; set; }
+            public string Index { get; set; }
+            public long TotalHits { get; set; }
+        }
+
+        public class V2SearchPackage
+        {
+            public V2SearchPackageRegistration PackageRegistration { get; set; }
+            public string Version { get; set; }
```
joelverhagen: Please only add properties you are using and validating against. This is unnecessary boiler plate right now. (right?)

src/NuGet.Services.EndToEnd/PushTests.cs
joelverhagen: Mark as `SemVer2Fact`

src/NuGet.Services.EndToEnd/PushTests.cs
```diff
@@ -55,5 +55,32 @@ public PushTests(PushedPackagesFixture pushedPackages, ITestOutputHelper logger)
...
+            var shouldBeEmptyV2 = await _clients.V3Search.SearchQuery($"q={package.Id}");
+            var shouldBeEmptyAutocomplete = await _clients.V3Search.Autocomplete($"q={package.Id}");
+            Assert.Equal(0, shouldBeEmptyV3.Data.Count);
+            Assert.Equal(0, shouldBeEmptyV2.Data.Count);
+            Assert.Equal(0, shouldBeEmptyAutocomplete.Data.Count);
+
+            var shouldNotBeEmptyV3 = await _clients.V3Search.Query($"q={package.Id}&semVerLevel=2.0.0");
```
joelverhagen: No need to test `/search/query` in my opinion. This is really an implementation detail and not a publicly used endpoint. You can push back if you want though.

Use `packageid:` for exact match.

`Assert.Equal(1, ...)` since we know the ID is unique.

Shouldn't `prerelease=true` be required?

src/NuGet.Services.EndToEnd/PushTests.cs
ryuyu: done

src/NuGet.Services.EndToEnd/PushTests.cs
ryuyu: True. Removed unnecessary waits.

src/NuGet.Services.EndToEnd/PushTests.cs
ryuyu: done

src/NuGet.Services.EndToEnd/PushTests.cs
ryuyu: I think this was only filtered anyway? Just bad naming on my part.

src/NuGet.Services.EndToEnd/PushTests.cs
ryuyu: i'm ok with forgoing /search/query testing. I added it becuase it was an endpoint we wanted to test.
And yes, prerelease flag should be set. Adding that too.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
ryuyu: Sure. Just thought i would flesh them out here so we don't have to go searching for it later, but i can remove it if you don't think it's useful.

src/NuGet.Services.EndToEnd/PushTests.cs
ryuyu: reorganized.

src/NuGet.Services.EndToEnd/PushTests.cs
```diff
@@ -55,5 +55,28 @@ public PushTests(PushedPackagesFixture pushedPackages, ITestOutputHelper logger)
...
+        public async Task NewSemVer2PackageIsFiltered()
+        {
+            // Arrange
+            var package = await _pushedPackages.PushAsync(PackageType.SemVer2Prerelease, _logger);
+
+            // Act
+            // Wait for package to become available
+            await _clients.V3Search.WaitForPackageAsync(package.Id, package.Version, _logger);
+
+            var shouldBeEmptyV3 = await _clients.V3Search.Query($"q={package.Id}");
```
joelverhagen: `&prerelease=true` here. The only different between "should be empty" and "should not be empty" is the `semVerLevel=2.0.0`.

src/NuGet.Services.EndToEnd/SearchResultTests.cs
```diff
@@ -0,0 +1,50 @@
+﻿using NuGet.Services.EndToEnd.Support;
```
joelverhagen: Copyright heading

Sort usings

src/NuGet.Services.EndToEnd/SearchResultTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        public SearchResultTests(ITestOutputHelper logger)
+        {
+            _clients = Clients.Initialize();
+            _logger = logger;
+        }
+        
+        [SemVer2Fact]
+        public async Task RegistrationValuesInResultMatchRequestSemVerLevel()
+        {
+            // Arrange
```
joelverhagen: Technically, we need to push both a semver 1 and semver 2 package here. They should be the same as the ones the other test use.

Imagine running this on a completely new environment.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -0,0 +1,253 @@
...
+            if (_testSettings.SearchBaseUrl != null)
+            {
+                searchBaseUrls.Add(_testSettings.SearchBaseUrl);
+            }
+            else
+            {
+                searchBaseUrls.AddRange(await _v3IndexClient.GetSearchBaseUrls());
+            }
+
+            var queryUrl = searchBaseUrls
```
joelverhagen: Hm... `searchBaseUrl` should be a parameter. All search services should be tested.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -0,0 +1,253 @@
...
+                string.Join(Environment.NewLine, v2SearchUrls.Select(u => $" - {u}")));
+
+            var tasks = v2SearchUrls
+                .Select(u => WaitForPackageAsync(u, id, version, logger))
+                .ToList();
+
+            await Task.WhenAll(tasks);
+        }
+        
+        public async Task<IReadOnlyList<string>> GetSearchBaseUrlsAsync()
```
joelverhagen: Call this to get the list of search base URLs to test. Then enumerate calling the `Query` method on each. You could add a helper method for this that returns a dictionary where the key is the search base URL and the value is the search result from that specific search service. Whichever is clearer for you.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
joelverhagen: I don't think it's useful until we have assertions for it. This sort of resolution makes sense for search functional tests but not e2e.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -0,0 +1,253 @@
...
+        private readonly TestSettings _testSettings;
+
+        public V2V3SearchClient(SimpleHttpClient httpClient, V3IndexClient v3IndexClient, TestSettings testSettings)
+        {
+            _httpClient = httpClient;
+            _v3IndexClient = v3IndexClient;
+            _testSettings = testSettings;
+        }
+
+        public async Task<V2SearchResponse> SearchQuery(string queryString)
```
joelverhagen: suffix async method names with `Async`, a la `SearchQueryAsync`.

joelverhagen: Actually this method is not used now right? `Async` the other ones tho.

src/NuGet.Services.EndToEnd/SearchResultTests.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            for (var i = 0; i < semVer1Query.Data.Count; i++)
+            {
+                Assert.False(semVer2RegistrationAddresses.Any(a => semVer1Query.Data[i].Registration.Contains(a)));
+                Assert.True(allRegistrationAddresses.Any(a => semVer1Query.Data[i].Registration.Contains(a)));
+            }
+
+            for (var i = 0; i < semVer2Query.Data.Count; i++)
+            {
+                Assert.True(semVer2RegistrationAddresses.Any(a => semVer2Query.Data[i].Registration.Contains(a)));
```
joelverhagen: There should be a symmetrical `Assert.False` here for `allRegistrationAddresses` right?

src/NuGet.Services.EndToEnd/SearchResultTests.cs
joelverhagen: To do this add `[Collection(nameof(PushedPackagesCollection))]` as a class-level attribute and `PushedPackagesFixture pushedPackages` as a ctor arg.

src/NuGet.Services.EndToEnd/SearchResultTests.cs
ryuyu: I thought allRegistrationAddresses included the semver2 registrations? In any case, since the semver2 set is a subset of total, asserting that the value is always one of the ones contained in semver2 set is a sufficient assertion.

joelverhagen: ah, nice!

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
ryuyu: Yes. I've "async"ed them all, but searchQuery will currently be unused.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
ryuyu: Added looping over the base urls to each test.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
ryuyu: Changed to pass in as a param.

src/NuGet.Services.EndToEnd/SearchResultTests.cs
ryuyu: Do i need to actually call the push to ensure they are pushed, or is it sufficient to add those?

src/NuGet.Services.EndToEnd/PushTests.cs
ryuyu: done

src/NuGet.Services.EndToEnd/PushTests.cs
```diff
@@ -55,5 +55,32 @@ public PushTests(PushedPackagesFixture pushedPackages, ITestOutputHelper logger)
...
+        [SemVer2Fact]
+        public async Task NewSemVer2PackageIsFiltered()
+        {
+            // Arrange
+            var package = await _pushedPackages.PushAsync(PackageType.SemVer2Prerelease, _logger);
+            var searchBaseAddresses = await _clients.V3Index.GetSearchBaseUrls();
+
+            // Act
+            // Wait for package to become available
+            await _clients.V3Search.WaitForPackageAsync(package.Id, package.Version, _logger);
```
joelverhagen: nit: this should be in the arrange area not act

PR #7, opened 2017-05-01
Add end-to-end test to verify unlisted packages in registration and search

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/3786.

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
@@ -21,12 +21,27 @@ namespace NuGet.Services.EndToEnd.Support
 
         public async Task PushAsync(Stream nupkgStream)
         {
-            var pushUrl = $"{_testSettings.GalleryBaseUrl}/api/v2/package";
+            var url = $"{_testSettings.GalleryBaseUrl}/api/v2/package";
             using (var httpClient = new HttpClient())
-            using (var request = new HttpRequestMessage(HttpMethod.Put, pushUrl))
+            using (var request = new HttpRequestMessage(HttpMethod.Put, url))
             {
                 request.Headers.Add("X-NuGet-ApiKey", _testSettings.ApiKey);
                 request.Content = new StreamContent(nupkgStream);
+
+                using (var response = await httpClient.SendAsync(request))
+                {
+                    response.EnsureSuccessStatusCode();
+                }
+            }
+        }
+
+        public async Task UnlistAsync(string id, string version)
+        {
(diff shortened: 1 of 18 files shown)


src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
```diff
@@ -34,5 +34,20 @@ public GalleryClient(TestSettings testSettings)
...
             }
         }
+
+        public async Task UnlistAsync(string id, string version)
+        {
+            var url = $"{_testSettings.GalleryBaseUrl}/api/v2/package/{id}/{version}";
+            using (var httpClient = new HttpClient())
+            using (var request = new HttpRequestMessage(HttpMethod.Delete, url))
+            {
+                request.Headers.Add("X-NuGet-ApiKey", _testSettings.ApiKey);
```
skofman1: const for X-NuGet-ApiKey

joelverhagen: Good idea, thanks.

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -8,6 +8,20 @@ namespace NuGet.Services.EndToEnd.Support
 {
     public class TestSettings
     {
+        public enum Mode
+        {
+            EnvironmentVariables,
```
skofman1: >EnvironmentVariables [](start = 12, length = 20)

consider renaming to "Local"

joelverhagen: I thought of this but decided against because local is confusing since that could suggest localhost gallery or search.

src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
joelverhagen: Fixed.

skofman1:
>             string apiKey,

Does it make sense to have API key as part of the test settings? You can't hard coded it bellow (and are forced to use an empty value), and there might be several API keys to be used in different scenarios.

---
Refers to: src/NuGet.Services.EndToEnd/Support/TestSettings.cs:29 in 03f417a. [](commit_id = 03f417a3da64efd8a18c8be49a9907893b5aadd7, deletion_comment = False)

joelverhagen:
> Does it make sense to have API key as part of the test settings? You can't hard coded it bellow (and are forced to use an empty value), and there might be several API keys to be used in different scenarios.

Spoke offline about this. We decided to add multiple API keys to the framework when the need arises. In general, this sort of edge case testing should be done in functional test projects for specific components (as NuGetGallery already has).

PR #13, opened 2017-06-14
E2E test coverage for autocomplete service hijack

xavierdecoster:
Fixes https://github.com/NuGet/NuGetGallery/issues/4119

### DIFF
--- a/src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
+++ b/src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
@@ -1,0 +1,97 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using NuGet.Services.EndToEnd.Support;
+using NuGet.Versioning;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace NuGet.Services.EndToEnd
+{
+    public class AutoCompleteResultTests
+        : IClassFixture<TrustedHttpsCertificatesFixture>
+    {
+        private readonly TestSettings _testSettings;
+        private readonly ITestOutputHelper _logger;
+        private readonly Clients _clients;
+
+        public AutoCompleteResultTests(ITestOutputHelper logger)
+        {
+            _clients = Clients.Initialize();
+            _testSettings = TestSettings.Create();
+            _logger = logger;
+        }
+
+        [Theory]
+        [InlineData(true, null)]
+        [InlineData(false, null)]
+        [InlineData(true, "2.0.0")]
+        [InlineData(false, "2.0.0")]
+        public async Task V2PackageVersionsAutocompleteResultsMatchV3Results(
+            bool includePrerelease,
+            string semVerLevel)
+        {
+            // Arrange
+            const string id = "EntityFramework";
+
+            var v2AutoCompletePackageVersionsEndpoint = $"{_testSettings.GalleryBaseUrl}/api/v2/package-versions";
+            var v2AutoCompletePackageVersionsQueryString = BuildV2QueryString(id, includePrerelease, semVerLevel);
+            var v2Url = v2AutoCompletePackageVersionsEndpoint + v2AutoCompletePackageVersionsQueryString;
+
+            var v3AutoCompletePackageVersionsEndpoints = await _clients.V2V3Search.GetAutoCompleteUrlsAsync();
+            var v3AutoCompletePackageVersionsQueryString = BuildV3QueryString($"id={id}", includePrerelease, semVerLevel);
+
+            // Act
+            var httpClient = new SimpleHttpClient();
+            var v2Response = await httpClient.GetJsonAsync<List<string>>(v2Url, _logger);
+
+            foreach (var v3AutoCompletePackageVersionsEndpoint in v3AutoCompletePackageVersionsEndpoints)
+            {
+                var v3Url = v3AutoCompletePackageVersionsEndpoint + v3AutoCompletePackageVersionsQueryString;
+                var v3Response = await httpClient.GetJsonAsync<List<string>>(v2Url, _logger);
+
+                // Assert
+                Assert.Equal(v2Response.Count, v3Response.Count);
+
+                for (int i = 0; i < v2Response.Count; i++)
+                {
+                    Assert.Equal(v2Response[i], v3Response[i]);
+                }
+            }
+        }
+
+        private static string BuildV2QueryString(string id, bool? includePrerelease, string semVerLevel = null)
+        {
+            var queryString = $"includePrerelease={includePrerelease ?? false}";
+
+            NuGetVersion semVerLevelVersion;
+            if (!string.IsNullOrEmpty(semVerLevel) && NuGetVersion.TryParse(semVerLevel, out semVerLevelVersion))
+            {
+                queryString += $"&semVerLevel={semVerLevel}";
+            }
+
+            return $"/{id}?" + queryString;
+        }
+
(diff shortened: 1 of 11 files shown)


src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            _clients = Clients.Initialize();
+            _testSettings = TestSettings.Create();
+            _logger = logger;
+        }
+
+        [Theory]
+        [InlineData(true, null)]
+        [InlineData(false, null)]
+        [InlineData(true, "2.0.0")]
+        [InlineData(false, "2.0.0")]
```
joelverhagen: `SemVer2Fact` on the second two

src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
```diff
@@ -0,0 +1,97 @@
...
+        public async Task V2PackageVersionsAutocompleteResultsMatchV3Results(
+            bool includePrerelease,
+            string semVerLevel)
+        {
+            // Arrange
+            const string id = "EntityFramework";
```
joelverhagen: End-to-end tests should not assume any specific packages existing that will never be pushed by the tests themselves.

src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            var v2AutoCompletePackageVersionsEndpoint = $"{_testSettings.GalleryBaseUrl}/api/v2/package-versions";
+            var v2AutoCompletePackageVersionsQueryString = BuildV2QueryString(id, includePrerelease, semVerLevel);
+            var v2Url = v2AutoCompletePackageVersionsEndpoint + v2AutoCompletePackageVersionsQueryString;
+
+            var v3AutoCompletePackageVersionsEndpoints = await _clients.V2V3Search.GetAutoCompleteUrlsAsync();
+            var v3AutoCompletePackageVersionsQueryString = BuildV3QueryString($"id={id}", includePrerelease, semVerLevel);
+
+            // Act
+            var httpClient = new SimpleHttpClient();
+            var v2Response = await httpClient.GetJsonAsync<List<string>>(v2Url, _logger);
```
joelverhagen: Please add helper methods to the minimal client implementations (`V2V3SearchClient` and `GalleryClient`).

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -35,6 +35,7 @@ public enum Mode
             IReadOnlyList<string> trustedHttpsCertificates,
             string apiKey,
             string searchBaseUrl,
+            string autoCompleteUrl,
```
joelverhagen: This can be implied by `searchBaseUrl` right? I don't see the value in the additional configuration.

src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            {
+                queryString += $"&semVerLevel={semVerLevel}";
+            }
+
+            return $"/{id}?" + queryString;
+        }
+
+        private static string BuildV3QueryString(string queryString, bool? includePrerelease, string semVerLevel = null)
+        {
+            queryString += $"&prerelease={includePrerelease ?? false}";
```
joelverhagen: Please use `Microsoft.AspNetCore.WebUtilities.QueryHelpers` for query string building.

src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            {
+                var v3Url = v3AutoCompletePackageVersionsEndpoint + v3AutoCompletePackageVersionsQueryString;
+                var v3Response = await httpClient.GetJsonAsync<List<string>>(v2Url, _logger);
+
+                // Assert
+                Assert.Equal(v2Response.Count, v3Response.Count);
+
+                for (int i = 0; i < v2Response.Count; i++)
+                {
+                    Assert.Equal(v2Response[i], v3Response[i]);
```
joelverhagen: `Assert.Equal` supports `IEnumerable<T>` right?

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
xavierdecoster: That is an assumption I'd not like to hard code in these tests. 

`SearchBaseUrl` is discovered by looking for the the `SearchGalleryQueryService/` resource type in `index.json`.

The proper way to discover the autocomplete endpoint URLs is to look for `AutocompleteServiceResourceType` in `index.json`, and not assume the URL value is the same for both resource types (even though these values are the same today).

src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
xavierdecoster: I guess you meant `SemVer2InlineData`, will fix.

src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
xavierdecoster: Agreed. Fixing to use the PushedPackagesFixture instead.

src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
joelverhagen: Oh, yeah. Thanks.

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
joelverhagen: In principle I totally agree with you. 

However when we want to run the search service against the staging slot, all we know is the base URL (by using the Azure management APIs to get a cloud service base URL). In that case we have to build the URL anyways. For these reason I unified the cases where a) we get the base URLs from v3/index.json and b) we get the staging base URL from Azure.

See https://github.com/NuGet/NuGet.Services.EndToEnd/blob/1cb7b82ad1932674a731e719912fd94ed1e13016/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs#L28-L48.

When the `searchBaseUrl` is configured in an E2E test run that means we are testing a staging slot of search service instead of the publicly visible search services. When will the `autoCompleteUrl` be specifified?

Also, nit, `autocompleteUrl` instead of `autoCompleteUrl` to align with the casing in v3/index.json.

xavierdecoster:
@joelverhagen can you please give it another pass?

src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
```diff
@@ -22,6 +26,24 @@ public GalleryClient(TestSettings testSettings)
             _testSettings = testSettings;
         }
 
+        public async Task<IList<string>> AutocompletePackageIdsAsync(string id, bool includePrerelease, string semVerLevel, ITestOutputHelper logger)
+        {
+            var serviceEndpoint = $"{_testSettings.GalleryBaseUrl}/api/v2/package-ids";
+            var uri = AppendAutocompletePackageIdsQueryString(serviceEndpoint, id, includePrerelease, semVerLevel);
+            var httpClient = new SimpleHttpClient();
```
joelverhagen: Can we make this passed into the ctor like the other clients?

xavierdecoster: Yep! Fixed in https://github.com/NuGet/NuGet.Services.EndToEnd/pull/13/commits/dadd1cf7ee93e779c3d45266f2d6ffbe16dee691

PR #15, opened 2017-06-28
Need retry logic for WaitForPackageAsync in E2E clients

chenriksson:
Multiple deployments of catalog2registration have failed due to SocketException at SimpleHttpClient.GetJsonAsync. We should add retry logic here.

See E2E builds: 10372, 10387, 10388

Opened issue at: https://github.com/NuGet/NuGetGallery/issues/4295

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/Clients/SimpleHttpClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/SimpleHttpClient.cs
@@ -18,6 +18,7 @@ namespace NuGet.Services.EndToEnd.Support
     public class SimpleHttpClient
     {
         private const int MaxRetries = 3;
+        private const int RetryDelayInMilliseconds = 500;
 
         public Task<T> GetJsonAsync<T>(string url, ITestOutputHelper logger)
         {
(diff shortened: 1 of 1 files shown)


src/NuGet.Services.EndToEnd/Support/Clients/SimpleHttpClient.cs
```diff
@@ -17,38 +17,59 @@ namespace NuGet.Services.EndToEnd.Support
...
+                                logger.WriteLine($" - URL: {url}{Environment.NewLine} - Response: {parsedJson.ToString(Formatting.None)}");
+                            }
 
-                using (var stream = await response.Content.ReadAsStreamAsync())
-                using (var streamReader = new StreamReader(stream))
+                            return JsonConvert.DeserializeObject<T>(json);
+                        }
+                    }
+                }
+                catch (Exception e)
```
joelverhagen: Could we retry on the minimal set of exceptions? e.g. `SocketException`?

chenriksson:
@joelverhagen going to close without merge after all. Per my email it seems like the failures may be due to a real regression rather than flakiness. I now agree with you... while the retry may improve test resilience, these timeouts may be necessary to detect perf regressions until we have proper performance tests.

PR #16, opened 2017-07-19
Shorten timeout of flat container and registration

loic-sharma:
catalog2dnx and catalog2registration jobs do not need a long timeout like catalog2lucene. This shortens the timeout so that the tests fail faster in the error case.

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/TestData.cs
+++ b/src/NuGet.Services.EndToEnd/Support/TestData.cs
@@ -12,6 +12,8 @@ namespace NuGet.Services.EndToEnd.Support
     public static class TestData
     {
         public static readonly NuGetFramework TargetFramework = NuGetFramework.Parse("net40");
+        public static readonly TimeSpan FlatContainerWaitDuration = TimeSpan.FromMinutes(10);
+        public static readonly TimeSpan RegistrationWaitDuration = TimeSpan.FromMinutes(10);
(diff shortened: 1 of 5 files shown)


src/NuGet.Services.EndToEnd/Support/TestData.cs
```diff
@@ -12,6 +12,8 @@ namespace NuGet.Services.EndToEnd.Support
     public static class TestData
     {
         public static readonly NuGetFramework TargetFramework = NuGetFramework.Parse("net40");
+        public static readonly TimeSpan FlatContainerWaitDuration = TimeSpan.FromMinutes(10);
+        public static readonly TimeSpan RegistrationWaitDuration = TimeSpan.FromMinutes(10);
         public static readonly TimeSpan V3WaitDuration = TimeSpan.FromMinutes(20);
```
joelverhagen: Can we rename this to "search wait duration"?

src/NuGet.Services.EndToEnd/Support/TestData.cs
```diff
@@ -12,6 +12,8 @@ namespace NuGet.Services.EndToEnd.Support
     public static class TestData
     {
         public static readonly NuGetFramework TargetFramework = NuGetFramework.Parse("net40");
+        public static readonly TimeSpan FlatContainerWaitDuration = TimeSpan.FromMinutes(10);
```
joelverhagen: Is 10 minutes too long? Could we get away with 5 minutes?

PR #17, opened 2017-07-21
Get search service instance count from Azure

skofman1:
After this PR I will add Gallery URL extraction at runtime and remove the setup script.

### DIFF
--- a/scripts/Configure-Environment.ps1
+++ b/scripts/Configure-Environment.ps1
@@ -4,10 +4,6 @@ param (
     [ValidateSet("Production", "Staging", "", IgnoreCase=$true)]
     [string]$GallerySlot,
     [string]$GalleryCloudServiceName,
-    [string]$SearchBaseUrl,
-    [ValidateSet("Production", "Staging", "", IgnoreCase=$true)]
-    [string]$SearchSlot,
-    [string]$SearchCloudServiceName,
     [string]$SubscriptionId,
     [string]$ApplicationId,
     [string]$TenantId,
@@ -74,15 +70,3 @@ Write-Host "Using the following GalleryBaseUrl: $GalleryBaseUrl"
 $env:GalleryBaseUrl = $GalleryBaseUrl
 Write-Host "##vso[task.setvariable variable=GalleryBaseUrl;]$GalleryBaseUrl"
 
-$SearchBaseUrl = Get-BaseUrl `
-    -ApplicationId $ApplicationId `
-    -CertificateThumbprint $CertificateThumbprint `
-    -SubscriptionId $SubscriptionId `
-    -TenantId $TenantId `
-    -BaseUrl $SearchBaseUrl `
-    -Slot $SearchSlot `
-    -CloudServiceName $SearchCloudServiceName
-
-Write-Host "Using the following SearchBaseUrl: $SearchBaseUrl" 
-$env:SearchBaseUrl = $SearchBaseUrl
-Write-Host "##vso[task.setvariable variable=SearchBaseUrl;]$SearchBaseUrl"
--- a/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
@@ -17,11 +17,6 @@ namespace NuGet.Services.EndToEnd.Support
 {
     public class V2V3SearchClient
     {
-        /// <summary>
-        /// In case cloud service information is not provided, this will be the fallback instance count.
-        /// </summary>
-        private const int DefaultServiceInstanceCount = 2;
-
         private readonly V3IndexClient _v3IndexClient;
         private readonly SimpleHttpClient _httpClient;
         private readonly TestSettings _testSettings;
@@ -164,7 +159,7 @@ namespace NuGet.Services.EndToEnd.Support
                 logger.WriteLine($"Extracting search services URLs from index.json: {_testSettings.V3IndexUrl}");
 
                 var urls = await _v3IndexClient.GetSearchBaseUrlsAsync();
-                return urls.Select(url => new SearchService { Uri = new Uri(url), InstanceCount = DefaultServiceInstanceCount }).ToList();
+                return urls.Select(url => new SearchService { Uri = new Uri(url), InstanceCount = _testSettings.SearchInstanceCount }).ToList();
             }
         }
 
--- a/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
@@ -27,16 +27,12 @@ namespace NuGet.Services.EndToEnd.Support
         private readonly TestSettings _testSettings;
         private readonly IAzureManagementAPIWrapper _azureManagementAPIWrapper; 
 
-        public V2V3SearchClient(SimpleHttpClient httpClient, V3IndexClient v3IndexClient, TestSettings testSettings)
+        public V2V3SearchClient(SimpleHttpClient httpClient, V3IndexClient v3IndexClient, TestSettings testSettings, IAzureManagementAPIWrapper azureManagementAPIWrapper)
         {
             _httpClient = httpClient;
             _v3IndexClient = v3IndexClient;
             _testSettings = testSettings;
-
-            if (testSettings.AzureManagementAPIWrapperConfiguration != null)
-            {
-                _azureManagementAPIWrapper = new AzureManagementAPIWrapper(testSettings.AzureManagementAPIWrapperConfiguration);
-            }
+            _azureManagementAPIWrapper = azureManagementAPIWrapper;
         }
 
         [Obsolete]
@@ -143,10 +139,15 @@ namespace NuGet.Services.EndToEnd.Support
             }
         }
 
-        public async Task<IReadOnlyList<SearchService>> GetSearchServicesAsync()
+        public async Task<IReadOnlyList<SearchService>> GetSearchServicesAsync(ITestOutputHelper logger)
         {
             if (_azureManagementAPIWrapper != null)
             {
+                logger.WriteLine($"Extracting search service properties from Azure. " +
+                    $"Subscription: {_testSettings.Subscription}," +
+                    $" Resource group: {_testSettings.SearchServiceResourceGroup}," +
+                    $" Service name: {_testSettings.SearchServiceName}");
+
                 string result = await _azureManagementAPIWrapper.GetCloudServicePropertiesAsync(
                                     _testSettings.Subscription,
                                     _testSettings.SearchServiceResourceGroup,
@@ -160,6 +161,8 @@ namespace NuGet.Services.EndToEnd.Support
             }
             else
             {
+                logger.WriteLine($"Extracting search services URLs from index.json: {_testSettings.V3IndexUrl}");
+
                 var urls = await _v3IndexClient.GetSearchBaseUrlsAsync();
                 return urls.Select(url => new SearchService { Uri = new Uri(url), InstanceCount = DefaultServiceInstanceCount }).ToList();
             }
@@ -195,7 +198,7 @@ namespace NuGet.Services.EndToEnd.Support
             string failureMessageFormat,
             ITestOutputHelper logger)
         {
-            var searchServices = await GetSearchServicesAsync();
+            var searchServices = await GetSearchServicesAsync(logger);
 
             var v2SearchUrls = searchServices
                 .SelectMany(service => GetSearchUrlsForPolling(service))
--- a/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
@@ -18,11 +18,6 @@ namespace NuGet.Services.EndToEnd.Support
     public class V2V3SearchClient
     {
         /// <summary>
-        /// To be used for <see cref="IAzureManagementAPIWrapper"/> request
-        /// </summary>
-        private const string ProductionSlot = "production";
-
-        /// <summary>
         /// In case cloud service information is not provided, this will be the fallback instance count.
         /// </summary>
         private const int DefaultServiceInstanceCount = 2;
@@ -156,7 +151,7 @@ namespace NuGet.Services.EndToEnd.Support
                                     _testSettings.Subscription,
                                     _testSettings.SearchServiceResourceGroup,
                                     _testSettings.SearchServiceName,
-                                    ProductionSlot,
+                                    _testSettings.SearchServiceSlot,
                                     CancellationToken.None);
 
                 var cloudService = AzureHelper.ParseCloudServiceProperties(result);
--- a/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
@@ -5,8 +5,10 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.WebUtilities;
+using NuGet.Services.AzureManagement;
 using NuGet.Versioning;
 using Xunit;
 using Xunit.Abstractions;
@@ -15,15 +17,31 @@ namespace NuGet.Services.EndToEnd.Support
 {
     public class V2V3SearchClient
     {
+        /// <summary>
+        /// To be used for <see cref="IAzureManagementAPIWrapper"/> request
+        /// </summary>
+        private const string ProductionSlot = "production";
+
+        /// <summary>
+        /// In case cloud service information is not provided, this will be the fallback instance count.
+        /// </summary>
+        private const int DefaultServiceInstanceCount = 2;
+
         private readonly V3IndexClient _v3IndexClient;
         private readonly SimpleHttpClient _httpClient;
         private readonly TestSettings _testSettings;
+        private readonly IAzureManagementAPIWrapper _azureManagementAPIWrapper; 
 
         public V2V3SearchClient(SimpleHttpClient httpClient, V3IndexClient v3IndexClient, TestSettings testSettings)
         {
             _httpClient = httpClient;
             _v3IndexClient = v3IndexClient;
             _testSettings = testSettings;
+
+            if (testSettings.AzureManagementAPIWrapperConfiguration != null)
+            {
+                _azureManagementAPIWrapper = new AzureManagementAPIWrapper(testSettings.AzureManagementAPIWrapperConfiguration);
+            }
         }
 
         [Obsolete]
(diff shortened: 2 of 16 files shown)


src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -15,15 +17,26 @@ namespace NuGet.Services.EndToEnd.Support
 {
     public class V2V3SearchClient
     {
+        /// <summary>
+        /// In case cloud service information is not provided, this will be the fallback instance count.
+        /// </summary>
+        private const int DefaultServiceInstanceCount = 2;
```
chenriksson: Is this the minValue for autoScale, to there are at least 2 instances always available?

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -133,19 +143,26 @@ private IEnumerable<string> GetSearchUrlsForPolling(string originalBaseUrl)
...
+
+                var cloudService = AzureHelper.ParseCloudServiceProperties(result);
+
+                return new List<SearchService>() { new SearchService { Uri = cloudService.Uri, InstanceCount = cloudService.InstanceCount } };
             }
             else
             {
-                searchBaseUrls.AddRange(await _v3IndexClient.GetSearchBaseUrlsAsync());
+                var urls = await _v3IndexClient.GetSearchBaseUrlsAsync();
+                return urls.Select(url => new SearchService { Uri = new Uri(url), InstanceCount = DefaultServiceInstanceCount }).ToList();
```
chenriksson: Should test output indicate whether AzureManagement API was used versus the default?

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -28,86 +31,148 @@ public enum Mode
...
+        /// </summary>
+        public static readonly Dictionary<Mode, TestSettings> DefaultTestSettings = new Dictionary<Mode, TestSettings>
+        {
+            {   Mode.Dev,
+                new TestSettings(
+                             DefaultAggressivePush,
+                             "https://dev.nugettest.org",
+                             "https://api.dev.nugettest.org/v3-index/index.json",
+                             new List<string>(),
+                             "fa133685-cfc3-43f3-81c7-0a36d0969987",
```
chenriksson: Should this be "API_KEY"? Is this a real key?

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -28,86 +31,148 @@ public enum Mode
...
+                return new CachingSecretReader(new KeyVaultReader(keyVaultConfiguration));
+            }
+
+            return new EmptySecretReader();
+        }
+
+        private static IAzureManagementAPIWrapperConfiguration GetAzureManagementConfiguration(ISecretReader secretReader)
+        {
+            AzureManagementAPIWrapperConfiguration azureConfig = null;
+            var clientId = secretReader.GetSecretAsync(EnvironmentSettings.AzureManagementAPIClientId).Result;
```
chenriksson: Assuming method isn't async since it's invoked from Lazy initializer?

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -15,15 +17,26 @@ namespace NuGet.Services.EndToEnd.Support
...
         private readonly TestSettings _testSettings;
+        private readonly IAzureManagementAPIWrapper _azureManagementAPIWrapper; 
 
         public V2V3SearchClient(SimpleHttpClient httpClient, V3IndexClient v3IndexClient, TestSettings testSettings)
         {
             _httpClient = httpClient;
             _v3IndexClient = v3IndexClient;
             _testSettings = testSettings;
+
+            if (testSettings.AzureManagementAPIWrapperConfiguration != null)
```
cristinamanum: Are the tests still functional if the testSettings.AzureManagementAPIWrapperConfiguration == null?

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -28,86 +31,148 @@ public enum Mode
...
+                            "https://api.int.nugettest.org/v3-index/index.json",
+                            new List<string>(),
+                            "API_KEY",
+                            semVer2Enabled: false)
+            },
+            {
+                Mode.Prod,
+                new TestSettings(
+                            DefaultAggressivePush,
+                            "https://www.nuget.org",
```
cristinamanum: Could these settings be configurable?

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -15,15 +17,26 @@ namespace NuGet.Services.EndToEnd.Support
...
+            {
+                _azureManagementAPIWrapper = new AzureManagementAPIWrapper(testSettings.AzureManagementAPIWrapperConfiguration);
```
joelverhagen: This should be a constructor param as well.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -133,19 +143,26 @@ private IEnumerable<string> GetSearchUrlsForPolling(string originalBaseUrl)
...
-        public async Task<IReadOnlyList<string>> GetSearchBaseUrlsAsync()
+        public async Task<IReadOnlyList<SearchService>> GetSearchServicesAsync()
         {
-            var searchBaseUrls = new List<string>();
-            if (_testSettings.SearchBaseUrl != null)
+            if (_azureManagementAPIWrapper != null)
             {
-                searchBaseUrls.Add(_testSettings.SearchBaseUrl);
+                string result = await _azureManagementAPIWrapper.GetCloudServicePropertiesAsync(
+                                    _testSettings.Subscription,
```
joelverhagen: nit indent only 4 spaces here

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
joelverhagen: In what case would this information not be provided? If the person running the test forgets a param? If we assume 2 and the load balancer sees 5 instances, then the test can fail.

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -28,86 +31,148 @@ public enum Mode
...
+
+                var keyVaultConfiguration = new KeyVaultConfiguration(vaultName, clientId, certificate);
+
+                return new CachingSecretReader(new KeyVaultReader(keyVaultConfiguration));
+            }
+
+            return new EmptySecretReader();
+        }
+
+        private static IAzureManagementAPIWrapperConfiguration GetAzureManagementConfiguration(ISecretReader secretReader)
```
joelverhagen: Helper methods should always be `async` make the sync caller call `.Result` or `.Wait`.

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
joelverhagen: You should probably reset this key 😄

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -28,86 +31,148 @@ public enum Mode
         /// </summary>
         public static bool DefaultAggressivePush => true;
 
+        /// <summary>
+        /// A list of default test settings per environment.
+        /// </summary>
+        public static readonly Dictionary<Mode, TestSettings> DefaultTestSettings = new Dictionary<Mode, TestSettings>
+        {
+            {   Mode.Dev,
```
joelverhagen: nit: new line before `Mode.Dev`

scripts/Configure-Environment.ps1
```diff
@@ -74,15 +70,3 @@ Write-Host "Using the following GalleryBaseUrl: $GalleryBaseUrl"
 $env:GalleryBaseUrl = $GalleryBaseUrl
```
joelverhagen: Either here or tracked by a backlog item -- we should eliminate the PowerShell script completely and fetch the staging base URL in C# as you have done with search.

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -28,86 +31,148 @@ public enum Mode
...
+        private static ISecretReader GetSecretReader()
+        {
+            string vaultName = EnvironmentSettings.KeyVaultName;
+
+            if (!string.IsNullOrEmpty(vaultName))
+            {
+                string clientId = EnvironmentSettings.KeyVaultClientId;
+                string certificateThumbprint = EnvironmentSettings.KeyVaultCertificateThumbprint;
+
+                var certificate = CertificateUtility.FindCertificateByThumbprint(StoreName.My, StoreLocation.LocalMachine, certificateThumbprint, validationRequired: false);
```
joelverhagen: nit: wrap long lines

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -28,86 +31,148 @@ public enum Mode
...
             }
         }
+
+        private static ISecretReader GetSecretReader()
+        {
+            string vaultName = EnvironmentSettings.KeyVaultName;
+
+            if (!string.IsNullOrEmpty(vaultName))
+            {
+                string clientId = EnvironmentSettings.KeyVaultClientId;
```
joelverhagen: var

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -28,86 +31,148 @@ public enum Mode
         /// </summary>
         public static bool DefaultAggressivePush => true;
 
+        /// <summary>
+        /// A list of default test settings per environment.
+        /// </summary>
+        public static readonly Dictionary<Mode, TestSettings> DefaultTestSettings = new Dictionary<Mode, TestSettings>
```
joelverhagen: Public static dictionary is a bad practice as they are mutable. This should be a method or an read only dictionary.

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
joelverhagen: There's also a unit test that asserts that this API key is not test. Are those passing?

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
skofman1: I removed InstanceCount from being provided in configuration. Now there are only 2 options:
1. Instance count is retrieved from Azure - will be the case for deployment scenarios
2. Default instance count (2) - for manual testing.

skofman1: @chenriksson - number of instances will never be below 2. This should be the config for auto scale, and also the depoyment config for search service in dev. If we have only a single instance we risk service unavailability.

joelverhagen: If we assume 2 and the actual count is lower, we have polling errors since some of the ports will not be listening. If we assume 2 and the actual count is higher, the `// Assert` step fail since the `// Act` step hit a search endpoint not polled.

The first case is pretty easy to identify but the second case would be very frustrating, even when testing locally.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
skofman1: Yes - if it's null we will fallback to get search service URI from index.json, and instance count will be default

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
skofman1: done

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
skofman1: done

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
skofman1: moved to a method

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
skofman1: LOL :) apparently not

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
skofman1: The default settings are only for local test purposes. 
When the tests run from VSTS they get config from  environment settings

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
skofman1: Yes

scripts/Configure-Environment.ps1
skofman1: Yes, that's what I wrote in the PR comments. I will do it as part of this work after the PR is completed

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
skofman1: yes

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
skofman1: The first case will never happened - if the instance count is less then 2 we have a serious problem with service config that needs to be fixed ASAP. 
Don't understand when second case is possible. Point me to the test that will fail.

joelverhagen: > The first case will never happened - if the instance count is less then 2 we have a serious problem with service config that needs to be fixed ASAP.

Okay, sounds good.

> Don't understand when second case is possible. Point me to the test that will fail.

Suppose you are running against a search service with 3 or more instances but you have defaulted to 2 in the test configuration.

See:
https://github.com/NuGet/NuGet.Services.EndToEnd/blob/9af5255d0610a97664f8feae5acca155c04010cd/src/NuGet.Services.EndToEnd/PushTests.cs#L53

This line should only pass if the package is available on _all_ search instances. However, if you have 2 configured and 3 in reality the aforementioned could pass when that third instance does not yet have the package.

Then this line could fail if the load balancer sends the request to that third instance:
https://github.com/NuGet/NuGet.Services.EndToEnd/blob/9af5255d0610a97664f8feae5acca155c04010cd/src/NuGet.Services.EndToEnd/PushTests.cs#L66

scripts/Configure-Environment.ps1
joelverhagen: Oops, missed that.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
skofman1: Got it. Added it back.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -322,7 +337,12 @@ public class AutocompleteResponse
             public string Index { get; set; }
             public long TotalHits { get; set; }
             public DateTimeOffset LastReopen { get; set; }
+        }
 
+        public class SearchService
```
joelverhagen: private class

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -27,87 +31,164 @@ public enum Mode
...
             string galleryBaseUrl,
             string v3IndexUrl,
+            int searchServiceInstanceCount,
             IReadOnlyList<string> trustedHttpsCertificates,
             string apiKey,
-            string searchBaseUrl,
             bool semVer2Enabled,
-            int searchInstanceCount)
+            IAzureManagementAPIWrapperConfiguration azureManagementAPIWrapperConfiguration = null,
+            string subscription = "",
```
joelverhagen: why not null?

joelverhagen: (instead of empty string)

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -133,19 +134,33 @@ private IEnumerable<string> GetSearchUrlsForPolling(string originalBaseUrl)
             }
         }
 
-        public async Task<IReadOnlyList<string>> GetSearchBaseUrlsAsync()
+        public async Task<IReadOnlyList<SearchService>> GetSearchServicesAsync(ITestOutputHelper logger)
         {
-            var searchBaseUrls = new List<string>();
-            if (_testSettings.SearchBaseUrl != null)
+            if (_azureManagementAPIWrapper != null)
```
joelverhagen: Previously `SearchBaseUrl` was only set when running end-to-end tests during the Search Service deployment pipeline. This was because we wanted to validate the staging slot. When running end-to-end tests during the gallery and V3 component deployments, the search base URLs came from index.json, meaning we would test all search endpoints (USNC and USSC).

It looks like we are changing this behavior. Am I correct in understanding that `_azureManagementAPIWrapper` will only ever be null when running locally?

Said another way, I think we need to get the instance count from the Azure management API no matter what (except maybe via an override for local testing). Whether the official end-to-end test run is testing the Staging slot or testing the search base URLs from index.json, we need the correct count.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
skofman1: It's used outside the class

joelverhagen: oh, my bad. let's unnest it and make it readonly then

skofman1: readonly?

joelverhagen: immutable class. meaning properties are just `{ get; }` and the values are initialized via ctor.

joelverhagen: I think the `set;` is only ever called when the instance is constructed, so we don't need a public setter.

skofman1: can we please focus on behavior issues in the code and not nits?

joelverhagen: Sounds fine!

PR #18, opened 2017-07-24
Search service polling should use HTTPS

skofman1:


src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -121,10 +121,10 @@ private IEnumerable<string> GetSearchUrlsForPolling(string originalBaseUrl)
         {
             for (var instanceIndex = 0; instanceIndex < _testSettings.SearchInstanceCount; instanceIndex++)
             {
-                var port = 8080 + instanceIndex;
+                var port = 44301 + instanceIndex;
```
scottbommarito: should we pass this in as a parameter instead?

also probably best as a constant rather than inline

PR #20, opened 2017-07-28
[ReleasePrep][2017.07.28]RI of dev into master

loic-sharma:
Includes #13, #16, and #19


loic-sharma:
Successful run from VSTS [here](https://nuget.visualstudio.com/NuGetBuild/_build/index?buildId=12254&_a=summary)

PR #23, opened 2017-09-22
Ensure the autocomplete tests wait for the package to become available.

xavierdecoster:
Fixes https://github.com/NuGet/Engineering/issues/747

### DIFF
--- a/src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
+++ b/src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
@@ -45,7 +45,8 @@ namespace NuGet.Services.EndToEnd
                 includePrerelease, 
                 semVerLevel, 
                 _logger);
-            
+
+            // Assert
             Assert.NotEmpty(v2Response);
 
             foreach (var searchService in await _clients.V2V3Search.GetSearchServicesAsync(_logger))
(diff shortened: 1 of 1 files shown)


src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
```diff
@@ -71,6 +74,9 @@ public AutocompleteResultTests(PushedPackagesFixture pushedPackages, ITestOutput
             // Arrange
             var package = await _pushedPackages.PrepareAsync(packageType, _logger);
 
+            // Wait for package to become available
```
joelverhagen: Can we also assert non-empty result of `v2Response` or `v3Response`?

xavierdecoster: Good one. Added https://github.com/NuGet/NuGet.Services.EndToEnd/pull/23/commits/955cf7d2d6a2674d1a186afc53d83d3b44fb9a91

src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
```diff
@@ -51,8 +56,9 @@ public AutocompleteResultTests(PushedPackagesFixture pushedPackages, ITestOutput
                     includePrerelease,
                     semVerLevel,
                     _logger);
-                
+
                 // Assert
```
joelverhagen: super-nit: move `// Assert` above `Assert.NotEmpty(v2Response);`

PR #34, opened 2018-03-06
Introduce a flag to disable running tests that check Gallery

skofman1:
We shouldn't run tests that hit Gallery for Asia deployments, since Gallery doesn't use Asia resources.

Part of this task: https://github.com/nuget/engineering/issues/1224

### DIFF
--- a/src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
+++ b/src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
@@ -23,7 +23,7 @@ namespace NuGet.Services.EndToEnd
             _logger = logger;
         }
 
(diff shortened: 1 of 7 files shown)


src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
```diff
@@ -23,6 +23,7 @@ public AutocompleteResultTests(PushedPackagesFixture pushedPackages, ITestOutput
             _logger = logger;
         }
 
+        //[GalleryTestTheory] Uncomment this line, and remove the Theory below once the test is fixed
```
cristinamanum: I think that the name suggests that it should be used for all the Gallery tests. Is this the intention?

PR #35, opened 2018-03-07
Fix search polling

ryuyu:
Addresses https://github.com/NuGet/NuGetGallery/issues/5407


src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -130,7 +130,7 @@ private IEnumerable<string> GetSearchUrlsForPolling(SearchServiceProperties sear
 
         public async Task<IReadOnlyList<SearchServiceProperties>> GetSearchServicesAsync(ITestOutputHelper logger)
         {
-            List<SearchServiceProperties> searchServices = null;
+            List<SearchServiceProperties> searchServices = new List<SearchServiceProperties>();
```
skofman1: >List<SearchServiceProperties [](start = 12, length = 28)

var

PR #39, opened 2018-03-30
[Package Signing] Add Signed Package Tests

loic-sharma:
See https://github.com/NuGet/NuGetGallery/issues/5497

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/EnvironmentSettings.cs
+++ b/src/NuGet.Services.EndToEnd/Support/EnvironmentSettings.cs
@@ -26,6 +26,11 @@ namespace NuGet.Services.EndToEnd.Support
         /// </summary>
         public static string ConfigurationName => GetEnvironmentVariable("ConfigurationName", required: true);
 
+        /// <summary>
+        /// The path to the signed package that should be used for tests.
+        /// </summary>
+        public static string SignedPackagePath = GetEnvironmentVariable("SignedPackagePath", required: false);
+
         private static string GetEnvironmentVariable(string key, bool required)
         {
             var output = Targets
--- a/src/NuGet.Services.EndToEnd/Support/Package.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Package.cs
@@ -63,6 +63,13 @@ namespace NuGet.Services.EndToEnd.Support
 
         public static Package SignedPackage()
         {
+            if (string.IsNullOrEmpty(EnvironmentSettings.SignedPackagePath) ||
+                !File.Exists(EnvironmentSettings.SignedPackagePath))
+            {
+                throw new InvalidOperationException(
+                    $"Environment variable '{EnvironmentSettings.SignedPackagePath}' must point to a valid path");
+            }
+
             string id;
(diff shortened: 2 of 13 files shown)


src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -254,6 +260,11 @@ private PackageToPrepare InitializePackage(PackageType packageType)
                         Package.Create(id, "1.0.0"),
                         unlist: true);
 
+                case PackageType.Signed:
+                    return new PackageToPrepare(
+                        Package.SignedPackage(),
+                        unlist: true);
```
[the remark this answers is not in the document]
joelverhagen: (this should be consistent with the other packages unrelated to unlist/relist scenarios

src/NuGet.Services.EndToEnd/Support/EnvironmentSettings.cs
```diff
@@ -26,6 +26,11 @@ public static class EnvironmentSettings
         /// </summary>
         public static string ConfigurationName => GetEnvironmentVariable("ConfigurationName", required: true);
 
+        /// <summary>
+        /// The path to the signed package that should be used for tests.
+        /// </summary>
+        public static string SignedPackagePath = GetEnvironmentVariable("SignedPackagePath", required: false);
```
skofman1: is this necessary because of the way we create the package? Since we are moving away from Environment Variables for config, it would be better if this could be configured in a json file, like everything else.

src/NuGet.Services.EndToEnd/Support/Package.cs
```diff
@@ -59,5 +60,38 @@ public static Package Create(PackageCreationContext context)
...
+
+        public static Package SignedPackage()
+        {
+            string id;
+            string normalizedVersion;
+            string fullVersion;
+
+            ReadOnlyCollection<byte> nupkgBytes;
+
+            using (var nupkgStream = File.Open(EnvironmentSettings.SignedPackagePath, FileMode.Open))
```
skofman1: if the path is not set this will throw. Check input and return null.

src/NuGet.Services.EndToEnd/Support/EnvironmentSettings.cs
loic-sharma: Yup, the VSTS build itself sets the environment variable if it has the `GenerateSignedPackage` step. If you don't have that step, the environment variable won't exist and the signed package tests won't run.

src/NuGet.Services.EndToEnd/Support/Package.cs
loic-sharma: Fixed. I made this method throw an exception that details that the environment variable must be set with a proper path as that'll be easier to debug.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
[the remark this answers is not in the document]
loic-sharma: Fixed

PR #43, opened 2018-06-20
E2E:  ensure Items properties are non-null

dtivel:
Resolve https://github.com/NuGet/Engineering/issues/1483.


src/NuGet.Services.EndToEnd/Support/Clients/RegistrationClient.cs
```diff
@@ -164,11 +164,21 @@ public RegistrationClient(SimpleHttpClient httpClient, V3IndexClient v3IndexClie
         private class RegistrationIndexResponse
         {
             public List<RegistrationPage> Items { get; set; }
+
+            public RegistrationIndexResponse()
+            {
+                Items = new List<RegistrationPage>();
```
agr: I would initialize property inline instead of adding a constructor, it is allowed since C# 6, IIRC.
```
public List<RegistrationPage> Items { get; set; } = new List<RegistrationPage>();
```

PR #45, opened 2018-06-21
Fix E2E test

dtivel:
Resolve https://github.com/NuGet/Engineering/issues/1485.


src/NuGet.Services.EndToEnd/Support/Clients/NuGetVersionConverter.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Newtonsoft.Json;
+using NuGet.Versioning;
+
+namespace NuGet.Services.EndToEnd.Support
+{
+    public sealed class NuGetVersionConverter : JsonConverter
```
scottbommarito: Any reason you can't use https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol/Converters/NuGetVersionConverter.cs directly? Are you avoiding taking an additional dependency?

PR #47, opened 2018-06-22
[Package Signing] Add tests for repository signing

loic-sharma:
Here is an example run: https://nuget.visualstudio.com/NuGetBuild/_build/results?buildId=34801&view=logs

Depends on: https://nuget.visualstudio.com/_git/NuGetBuild/pullrequest/543?_a=overview

Fixes https://github.com/NuGet/Engineering/issues/1205
Fixes https://github.com/NuGet/Engineering/issues/1206

### DIFF
--- a/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
+++ b/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
@@ -95,10 +95,7 @@
     <Content Include="Support\NuGetExe\nuget.4.1.0.exe">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
-    <Content Include="Support\NuGetExe\nuget.4.3.0-preview1-4044.exe">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </Content>
-    <Content Include="Support\NuGetExe\nuget.4.6.1.exe">
+    <Content Include="Support\NuGetExe\nuget.4.7.0.exe">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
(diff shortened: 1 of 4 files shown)


loic-sharma:
🔔 Ready for review! I verified this on DEV.

src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
```diff
@@ -101,6 +101,9 @@
     <Content Include="Support\NuGetExe\nuget.4.6.1.exe">
```
chenriksson: Is `nuget.4.6.1.exe` still used elsewhere? Does it need to be?

loic-sharma: Good question. The pattern has been to keep the nuget client around, but I'm not sure why. I'll do some digging.

loic-sharma: I did a scan through, I couldn't find anything using the previous clients. I went ahead and remove all unnecessary clients.

src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
```diff
@@ -95,10 +95,7 @@
     <Content Include="Support\NuGetExe\nuget.4.1.0.exe">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
-    <Content Include="Support\NuGetExe\nuget.4.3.0-preview1-4044.exe">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </Content>
-    <Content Include="Support\NuGetExe\nuget.4.6.1.exe">
+    <Content Include="Support\NuGetExe\nuget.4.7.0.exe">
```
scottbommarito: Out of curiosity, why are we keeping `4.1.0` but replacing the other versions?

loic-sharma: 4.1.0 is still used by tests due to a behavior change in the client.

loic-sharma:
Here is an example run: https://nuget.visualstudio.com/NuGetBuild/_build/results?buildId=34801&view=logs

Depends on: https://nuget.visualstudio.com/_git/NuGetBuild/pullrequest/543?_a=overview

Fixes https://github.com/NuGet/Engineering/issues/1205
Fixes https://github.com/NuGet/Engineering/issues/1206

PR #55, opened 2018-08-31
Log entire response headers and body when there is a non-success status code

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/6340

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/HttpResponseMessageExtensions.cs
+++ b/src/NuGet.Services.EndToEnd/Support/HttpResponseMessageExtensions.cs
@@ -1,0 +1,83 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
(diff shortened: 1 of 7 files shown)


src/NuGet.Services.EndToEnd/Support/HttpResponseMessageExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Threading.Tasks;
+using Xunit.Abstractions;
+
+namespace NuGet.Services.EndToEnd.Support
+{
+    public static class HttpResponseMessageExtensions
+    {
+        private const int LogBoundaryLength = 60;
+
+        public static async Task EnsureSuccessStatusCodeAndLogAsync(
```
loic-sharma: Should the name be `EnsureSuccessStatusCodeOrLogAsync` as this only logs if the status code isn't a success?

joelverhagen: I thought about that, but it is "ensuring" in either case, so it's not really an exclusive "or"... logical "or" perhaps... 😕....
Sure!

loic-sharma: Yeah.. Oh well, it's not a big deal. Both names work :)

PR #57, opened 2018-09-12
Fix null reference exception when non-success status code is returned when polling

joelverhagen:
Turns out the logger was null sometimes. I changed that.

This was a build run against DEV (no 500s):
https://nuget.visualstudio.com/NuGetBuild/_build/results?buildId=40178

This was a build run against INT (some 500s, but no null refs):
https://nuget.visualstudio.com/NuGetBuild/_build/results?buildId=40186


src/NuGet.Services.EndToEnd/Support/Clients/SimpleHttpClient.cs
```diff
@@ -35,10 +35,15 @@ internal SimpleHttpClient()
...
         public async Task<T> GetJsonAsync<T>(string url, ITestOutputHelper logger)
         {
-            return await GetJsonAsync<T>(url, allowNotFound: false, logger: logger);
+            return await GetJsonAsync<T>(url, allowNotFound: false, logResponseBody: true, logger: logger);
         }
 
         public async Task<T> GetJsonAsync<T>(string url, bool allowNotFound, ITestOutputHelper logger)
+        {
+            return await GetJsonAsync<T>(url, allowNotFound, logResponseBody: true, logger: logger);
+        }
```
chenriksson: How did you determine which json responses we should log?

joelverhagen: Polling requests do not log their response.

Everything else does.

PR #58, opened 2018-09-17
Retry on 5XX when polling flat container and registration blob storage

joelverhagen:
Progress on https://github.com/NuGet/NuGetGallery/issues/6103.

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/Clients/FlatContainerClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/FlatContainerClient.cs
@@ -5,6 +5,8 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Net;
+using System.Net.Sockets;
 using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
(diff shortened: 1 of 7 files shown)


joelverhagen:
E2E test run:
https://nuget.visualstudio.com/NuGetBuild/_build/results?buildId=40456&view=logs

1 500 from flat container. 1 from registration.

src/NuGet.Services.EndToEnd/Support/Clients/FlatContainerClient.cs
```diff
@@ -32,24 +34,37 @@ public FlatContainerClient(SimpleHttpClient httpClient, V3IndexClient v3IndexCli
...
 
-            var tasks = baseUrls
-                .Select(u => WaitForPackageAsync(u, id, version, logger))
-                .ToList();
+                    await Task.WhenAll(tasks);
+                },
+                ex => ex.HasTypeOrInnerType<SocketException>()
+                   || ex.HasTypeOrInnerType<TaskCanceledException>()
+                   || (ex.HasTypeOrInnerType<HttpRequestMessageException>(out var hre)
+                       && hre.StatusCode >= HttpStatusCode.InternalServerError),
```
dtivel: Should not retry 501 (`NotImplemented`) or 505 (`HttpVersionNotSupported`).

joelverhagen: If blob storage gives us these, what other action do we have but to retry? This is what the client does.

PR #59, opened 2018-09-24
Either test the staging search slot OR all production slots (never both)

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/1661.

I also need to remove the production slots from E2E configuration files. This will be a separate PR.

Also, make it so `SkipGalleryTests` covers all the things it should. V2 restore tests should be skipped.

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
@@ -30,6 +30,13 @@ namespace NuGet.Services.EndToEnd.Support
         private Uri _galleryUrl = null;
         private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1);
 
+        public GalleryClient(SimpleHttpClient httpClient, TestSettings testSettings, IRetryingAzureManagementAPIWrapper azureManagementAPIWrapper)
+        {
+            _httpClient = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
+            _testSettings = testSettings ?? throw new ArgumentNullException(nameof(testSettings));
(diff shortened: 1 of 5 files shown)


src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
```diff
@@ -30,6 +30,13 @@ public class GalleryClient : IGalleryClient
         private Uri _galleryUrl = null;
         private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1);
 
+        public GalleryClient(SimpleHttpClient httpClient, TestSettings testSettings, IRetryingAzureManagementAPIWrapper azureManagementAPIWrapper)
```
skofman1: null checks

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -172,8 +172,7 @@ public async Task<IReadOnlyList<SearchServiceProperties>> GetSearchServicesAsync
                         searchServices.Add(await GetSearchServiceFromAzureAsync(mappedService, logger));
                     }
                 }
-
-                if (_testSettings.SearchServiceConfiguration.SingleSearchService != null)
+                else
```
chenriksson: Does `_testSettings.SearchServiceConfiguration.IndexJsonMappedSearchServices == null` indicate single search? If so, is the `SingleSearchService` config unnecessary?

joelverhagen: This is actually the meat of the change. I intentionally changed the configuration so that you either test the production slots or the staging slot. Never both.

PR #62, opened 2018-11-15
When building with private build tools, copy over configuration file from private build tools

scottbommarito:
https://github.com/NuGet/Engineering/issues/1826


build.ps1
```diff
@@ -42,7 +42,25 @@ Trace-Log "Build #$BuildNumber started at $startTime"
...
+Invoke-BuildStep 'Getting private build tools' { 
+        Install-PrivateBuildTools 
+
+        if (-Not (Test-Path $PrivateRoot)) {
+            return
+        }
+
+        # Copy over the configuration file from the private build tools.
+        $SourcesDirectory = $env:Build_SourcesDirectory
+        $configFileName = "$env:ConfigurationName.json"
```
loic-sharma: String formatting like this scares me in PowerShell. Could we do `"$(env:ConfigurationName) abc"` instead?

scottbommarito: Using `$($env:ConfigurationName)` because `$(env:ConfigurationName)` doesn't work

PR #63, opened 2018-11-30
Merge E2E changes on master to dev

scottbommarito:
Lots of changes haven't been merged in.


joelverhagen:
Why do we need a `dev` branch at all?

scottbommarito:
@joelverhagen - it's for if someone wants to make a change without affecting E2E tests for deployments

joelverhagen:
When have we wanted to do this? Why won't a private branch suffice?

PR #65, opened 2018-12-14
Add Symbols end to end tests

shishirx34:
This PR adds end to end test for symbols. Addresses https://github.com/NuGet/NuGetGallery/issues/5520

The test basically, builds sample template project and creates nupkg and snupkg with dlls and pdbs respectively. Pushes this to gallery and then listens to the symbol server endpoint for the availability of the PDB.

Had to update the target framework to net46 due to updated client dependencies which were required to allow symbols validations using client code.

Edit: I will add more tests to this as appropriate.

### DIFF
--- a/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
+++ b/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
@@ -41,8 +41,12 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="AutocompleteResultTests.cs" />
+    <Compile Include="Support\PackageProperties.cs" />
+    <Compile Include="Support\Utilities\PortableMetadataReader.cs" />
+    <Compile Include="SymbolsPackageTests.cs" />
     <Compile Include="Support\Clients\ClientHelper.cs" />
     <Compile Include="Support\Clients\IRetryingAzureManagementAPIWrapper.cs" />
+    <Compile Include="Support\Clients\DotNetExeClient.cs" />
     <Compile Include="Support\Clients\RetryingAzureManagementAPIWrapper.cs" />
     <Compile Include="Support\Clients\SearchServiceProperties.cs" />
     <Compile Include="Support\Clients\V3IndexClient.cs" />
@@ -107,6 +111,15 @@
     <Content Include="Support\NuGetExe\nuget.4.9.1.exe">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="Support\TestData\E2E.TestPortableSymbols\E2E.TestPortableSymbols.csproj">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="Support\TestData\E2E.TestPortableSymbols\TestMath.cs">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="Support\TestData\E2E.TestPortableSymbols\Properties\AssemblyInfo.cs">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <ItemGroup>
@@ -124,6 +137,7 @@
     <ExternalConfig Include="ExternalConfig/*.*">
     </ExternalConfig>
   </ItemGroup>
+  <ItemGroup />
   <Target Name="AfterBuild">
     <Copy SourceFiles="@(ExternalConfig)" DestinationFolder="$(TargetDir)Config" />
   </Target>
--- a/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
+++ b/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
@@ -100,9 +100,7 @@
     <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
   </ItemGroup>
   <ItemGroup>
-    <None Include="App.config">
-      <SubType>Designer</SubType>
-    </None>
+    <None Include="App.config" />
     <None Include="project.json" />
   </ItemGroup>
   <ItemGroup>
@@ -141,7 +139,6 @@
     <ExternalConfig Include="ExternalConfig/*.*">
     </ExternalConfig>
   </ItemGroup>
-  <ItemGroup />
   <Target Name="AfterBuild">
     <Copy SourceFiles="@(ExternalConfig)" DestinationFolder="$(TargetDir)Config" />
   </Target>
--- a/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
+++ b/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
@@ -9,8 +9,9 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>NuGet.Services.EndToEnd</RootNamespace>
     <AssemblyName>NuGet.Services.EndToEnd</AssemblyName>
-    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
+    <TargetFrameworkProfile />
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>true</DebugSymbols>
@@ -98,7 +99,9 @@
     <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
   </ItemGroup>
   <ItemGroup>
-    <None Include="App.config" />
+    <None Include="App.config">
+      <SubType>Designer</SubType>
+    </None>
     <None Include="project.json" />
   </ItemGroup>
   <ItemGroup>
--- a/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
+++ b/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
@@ -42,6 +42,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="AutocompleteResultTests.cs" />
+    <Compile Include="Support\Clients\SymbolServerClient.cs" />
     <Compile Include="Support\PackageProperties.cs" />
     <Compile Include="Support\Utilities\PortableMetadataReader.cs" />
     <Compile Include="SymbolsPackageTests.cs" />
--- a/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
+++ b/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
@@ -9,7 +9,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>NuGet.Services.EndToEnd</RootNamespace>
     <AssemblyName>NuGet.Services.EndToEnd</AssemblyName>
-    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <TargetFrameworkProfile />
   </PropertyGroup>
@@ -100,7 +100,9 @@
     <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
   </ItemGroup>
   <ItemGroup>
-    <None Include="App.config" />
+    <None Include="App.config">
+      <SubType>Designer</SubType>
+    </None>
     <None Include="project.json" />
   </ItemGroup>
   <ItemGroup>
--- a/src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
@@ -108,10 +108,13 @@ namespace NuGet.Services.EndToEnd.Support
             return await _httpClient.GetJsonAsync<List<string>>(uri, logger);
         }
 
-        public async Task PushAsync(Stream nupkgStream, ITestOutputHelper logger)
+        public async Task PushAsync(Stream nupkgStream, ITestOutputHelper logger, bool isSymbolsPackage = false)
         {
             var galleryEndpoint = await GetGalleryUrlAsync(logger);
-            var url = $"{galleryEndpoint}/api/v2/package";
+            var url = isSymbolsPackage
+                ? $"{galleryEndpoint}/api/v2/symbolpackage"
+                : $"{galleryEndpoint}/api/v2/package";
+
             using (var httpClient = new HttpClient())
             using (var request = new HttpRequestMessage(HttpMethod.Put, url))
             {
--- a/src/NuGet.Services.EndToEnd/Support/Clients/SymbolServerClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/SymbolServerClient.cs
@@ -74,7 +74,9 @@ namespace NuGet.Services.EndToEnd.Support
             {
                 using (var httpClient = new HttpClient())
                 {
-                    // Add Checksum headers for fetching the pdbs from symbol server.
+                    // Add Checksum headers for fetching the pdbs from symbol server. This header is used
+                    // by the symbol server to validate the request(provided by supported clients) for 
+                    // portable PDBs.
                     httpClient.DefaultRequestHeaders.Add(SymbolServerHeaderKey, SymbolServerHeaderValue);
                     using (var response = await httpClient.GetAsync(symbolFileEndpoint))
                     {
--- a/src/NuGet.Services.EndToEnd/Support/Clients/SymbolServerClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/SymbolServerClient.cs
@@ -43,8 +43,6 @@ namespace NuGet.Services.EndToEnd.Support
             string failureMessageFormat,
             ITestOutputHelper logger)
         {
-            // We perform the retry at this level so that we can re-fetch the list of search service instances from
-            // Azure Management API. This list can change during scale-up or scale-down events.
             return RetryUtility.ExecuteWithRetry(
                 async () =>
                 {
--- a/src/NuGet.Services.EndToEnd/Support/Clients/SymbolServerClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/SymbolServerClient.cs
@@ -17,7 +17,6 @@ namespace NuGet.Services.EndToEnd.Support
     public class SymbolServerClient
     {
         private readonly TestSettings _testSettings;
-        private readonly string _symbolServerEndpointTemplate = "https://{0}/download/symbols/{1}";
         private static readonly string SymbolServerHeaderKey = "SymbolChecksum";
         private static readonly string SymbolServerHeaderValue = "1";
 
@@ -70,11 +69,7 @@ namespace NuGet.Services.EndToEnd.Support
             await Task.Yield();
 
             var complete = false;
-            var symbolFileEndpoint = new Uri(
-                string.Format(
-                    _symbolServerEndpointTemplate,
-                    _testSettings.SymbolServerDomain,
-                    indexedFileName));
+            var symbolFileEndpoint = new Uri(string.Format(_testSettings.SymbolServerUrlTemplate, indexedFileName));
 
             var duration = Stopwatch.StartNew();
             while (!complete && duration.Elapsed < TestData.SymbolsWaitDuration)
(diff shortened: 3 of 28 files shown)


src/NuGet.Services.EndToEnd/Support/Package.cs
```diff
@@ -100,5 +135,19 @@ public static Package SignedPackage()
...
             return new Package(id, normalizedVersion, fullVersion, nupkgBytes);
         }
+
+        private static MemoryStream GetMemoryStreamForFile(string filename)
+        {
+            var memoryStream = new MemoryStream();
+            using (FileStream fileStream = File.OpenRead(filename))
+            {
+                memoryStream.SetLength(fileStream.Length);
+                fileStream.Read(memoryStream.GetBuffer(), 0, (int)fileStream.Length);
```
loic-sharma: Can you use [`Stream.CopyTo`]([https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.copyto?view=netframework-4.7.2#System_IO_Stream_CopyTo_System_IO_Stream](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.copyto?view=netframework-4.7.2#System_IO_Stream_CopyTo_System_IO_Stream)_) instead?

src/NuGet.Services.EndToEnd/Support/TestData.cs
```diff
@@ -16,32 +17,66 @@ public static class TestData
         public static readonly TimeSpan RegistrationWaitDuration = TimeSpan.FromMinutes(20);
         public static readonly TimeSpan SearchWaitDuration = TimeSpan.FromMinutes(35);
         public static readonly TimeSpan V3SleepDuration = TimeSpan.FromSeconds(5);
+        public static readonly TimeSpan SymbolsWaitDuration = TimeSpan.FromMinutes(20);
+        public static readonly TimeSpan SymbolsSleepDuration = TimeSpan.FromSeconds(5);
 
         public static Stream BuildPackageStream(PackageCreationContext context)
+        {
+            var files = new List<PhysicalPackageFile>() {
```
loic-sharma: > { [](start = 55, length = 2)

New line for `{`. We should add `.editorconfig` files to our projects...

src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
```diff
@@ -124,6 +141,7 @@
     <ExternalConfig Include="ExternalConfig/*.*">
     </ExternalConfig>
   </ItemGroup>
+  <ItemGroup />
```
joelverhagen: remove

src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
```diff
@@ -108,10 +108,13 @@ public async Task<IList<string>> AutocompletePackageVersionsAsync(string id, boo
             return await _httpClient.GetJsonAsync<List<string>>(uri, logger);
         }
 
-        public async Task PushAsync(Stream nupkgStream, ITestOutputHelper logger)
+        public async Task PushAsync(Stream nupkgStream, ITestOutputHelper logger, bool isSymbolsPackage = false)
         {
             var galleryEndpoint = await GetGalleryUrlAsync(logger);
-            var url = $"{galleryEndpoint}/api/v2/package";
+            var url = isSymbolsPackage
+                ? $"{galleryEndpoint}/api/v2/symbolpackage"
```
loic-sharma: Should we look up the push URLs from the service index? This is what the client does, it'd be a "stronger" test.

src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
```diff
@@ -108,10 +108,13 @@ public async Task<IList<string>> AutocompletePackageVersionsAsync(string id, boo
             return await _httpClient.GetJsonAsync<List<string>>(uri, logger);
         }
 
-        public async Task PushAsync(Stream nupkgStream, ITestOutputHelper logger)
+        public async Task PushAsync(Stream nupkgStream, ITestOutputHelper logger, bool isSymbolsPackage = false)
```
joelverhagen: Interface should match implementation

joelverhagen: (there is no default on the interface)

src/NuGet.Services.EndToEnd/Support/Clients/SymbolServerClient.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                using (var httpClient = new HttpClient())
+                {
+                    // Add Checksum headers for fetching the pdbs from symbol server.
+                    httpClient.DefaultRequestHeaders.Add(SymbolServerHeaderKey, SymbolServerHeaderValue);
+                    using (var response = await httpClient.GetAsync(symbolFileEndpoint))
+                    {
+                        complete = response.StatusCode == HttpStatusCode.OK;
+
+                        if (!complete && duration.Elapsed + TestData.SymbolsSleepDuration < TestData.SymbolsWaitDuration)
```
joelverhagen: this delay should be outside of the usings

src/NuGet.Services.EndToEnd/Support/Clients/SymbolServerClient.cs
```diff
@@ -0,0 +1,95 @@
...
+            var complete = false;
+            var symbolFileEndpoint = new Uri(string.Format(_testSettings.SymbolServerUrlTemplate, indexedFileName));
+
+            var duration = Stopwatch.StartNew();
+            while (!complete && duration.Elapsed < TestData.SymbolsWaitDuration)
+            {
+                using (var httpClient = new HttpClient())
+                {
+                    // Add Checksum headers for fetching the pdbs from symbol server.
+                    httpClient.DefaultRequestHeaders.Add(SymbolServerHeaderKey, SymbolServerHeaderValue);
```
loic-sharma: >SymbolServerHeaderValue [](start = 80, length = 23)

I'm not familiar with this header. Does this mean the symbol server should do checksum verification? Could you expand the comment and maybe link some documentation?

src/NuGet.Services.EndToEnd/Support/PackageProperties.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace NuGet.Services.EndToEnd.Support
+{
+    public class PackageProperties
+    {
+        public bool IsSymbolsPackage { get; }
+
+        public HashSet<string> IndexedFiles { get; }
+
+        public PackageProperties() { }
```
joelverhagen: Shouldn't this default `IndexedFiles` to an empty set?

src/NuGet.Services.EndToEnd/Support/PackageType.cs
```diff
@@ -19,5 +19,6 @@ public enum PackageType
         SemVer2StableMetadataUnlisted,
         FullValidation,
         Signed,
+        SymbolsPackage
```
joelverhagen: nit: trailing comma

src/NuGet.Services.EndToEnd/Support/PackageProperties.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public PackageProperties(bool isSymbolsPackage, HashSet<string> indexedFiles)
+        {
+            IsSymbolsPackage = isSymbolsPackage;
+            IndexedFiles = indexedFiles;
+        }
+
+        public static PackageProperties Default()
```
joelverhagen: Why have this if the default ctor is public?

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -96,21 +109,21 @@ private async Task<Package> PreparePackagesAsync(PackageType requestedPackageTyp
...
 
                 // Push all of the package types that have not been pushed yet.
-                var pushTasks = new Dictionary<PackageType, Task<Package>>();
+                var pushTasks = new Dictionary<PackageType, Task<List<Package>>>();
                 lock (_packagesLock)
                 {
                     foreach (var packageType in packageTypes)
                     {
-                        if (!_packages.TryGetValue(packageType, out Package package))
+                        if (!_packages.TryGetValue(packageType, out List<Package> package))
```
joelverhagen: `IReadOnlyList<Package>`

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -234,40 +282,132 @@ private PackageToPrepare InitializePackage(PackageType packageType)
...
+                    throw new Exception($"Error building symbols package! Error: {buildCommandResult.Error} Output: {buildCommandResult.Output}");
+                }
+
+                // Build nupkg and snupkg from appropriate files
+                var buildOutput = Path.Combine(testDirectory.FullPath, "bin", "Debug");
+                var dllFiles = Directory.GetFiles(buildOutput, "*.dll");
+                var pdbFiles = Directory.GetFiles(buildOutput, "*.pdb");
+                if (dllFiles.Count() == 0 || pdbFiles.Count() == 0)
+                {
+                    throw new Exception($"Symbols project build failed to create DLL or PDBs");
```
joelverhagen: Don't throw `Exception`. `InvalidOperationException` is better in this case.

An `Exception` cannot be caught. in a useful way.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -234,40 +282,132 @@ private PackageToPrepare InitializePackage(PackageType packageType)
...
+
+        private async Task<List<PackageToPrepare>> PrepareSymbolsPackageAsync(string id, string version, ITestOutputHelper logger)
+        {
+            using (var testDirectory = TestDirectory.Create())
+            {
+                // Copy the symbols project from template, also set the ID, Version appropriately in the properties.
+                var projectPath = CopySymbolsProjectFromTemplate(id, version, testDirectory.FullPath);
+
+                // Build the symbols project with the DotNet
+                var buildCommandResult = await DotNetExeClient.BuildProject(projectPath, logger);
```
joelverhagen: `BuildProjectAsync`

src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
```diff
@@ -9,8 +9,9 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>NuGet.Services.EndToEnd</RootNamespace>
     <AssemblyName>NuGet.Services.EndToEnd</AssemblyName>
-    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
```
joelverhagen: We moved to 4.6.2 most places. Can we do that?

src/NuGet.Services.EndToEnd/Support/Clients/SymbolServerClient.cs
shishirx34: I'll try to find documentation, not sure if there is. But this is the format for fetching portable pdbs from symbol server. I'll check with @cristinamanum.

src/NuGet.Services.EndToEnd/Support/Utilities/PortableMetadataReader.cs
```diff
@@ -0,0 +1,33 @@
...
+                throw new ArgumentNullException(nameof(pdbFullPath));
+            }
+
+            using (var stream = File.OpenRead(pdbFullPath))
+            using (var pdbReaderProvider = MetadataReaderProvider.FromPortablePdbStream(stream, MetadataStreamOptions.LeaveOpen))
+            {
+                var pdbReader = pdbReaderProvider.GetMetadataReader();
+                var pdbSignature = new BlobContentId(pdbReader.DebugMetadataHeader.Id).Guid;
+                var pdbFileName = Path.GetFileName(pdbFullPath).ToLowerInvariant();
+                var pdbAge = "FFFFFFFF";
```
loic-sharma: >FFFFFFFF [](start = 30, length = 8)

This should be "ffffffff" I believe

src/NuGet.Services.EndToEnd/Support/TestData/E2E.TestPortableSymbols/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("{0}")]
+[assembly: AssemblyDescription("")]
```
joelverhagen: Nit remove as unnecessary stuff as possible.

src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
shishirx34: I agree, but I would rather not make unrelated changes in this PR.

src/NuGet.Services.EndToEnd/Support/Utilities/PortableMetadataReader.cs
loic-sharma: Here's the latest implementation on BaGet: [https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.Core/Services/SymbolIndexingService.cs#L151](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.Core/Services/SymbolIndexingService.cs#L151)

I forget offhand what was broken with the uppercasing

---
In reply to: [241852502](https://github.com/NuGet/NuGet.Services.EndToEnd/pull/65#discussion_r241852502) [](ancestors = 241852502)

src/NuGet.Services.EndToEnd/Support/Utilities/PortableMetadataReader.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.IO;
+using System.Reflection.Metadata;
+
+namespace NuGet.Services.EndToEnd.Support.Utilities
+{
+    public static class PortableMetadataReader
+    {
+        public static string GetIndex(string pdbFullPath)
+        {
+            var indexingPath = string.Empty;
```
joelverhagen: nit set this after argument validation

src/NuGet.Services.EndToEnd/Support/Utilities/PortableMetadataReader.cs
```diff
@@ -0,0 +1,33 @@
...
+            }
+
+            using (var stream = File.OpenRead(pdbFullPath))
+            using (var pdbReaderProvider = MetadataReaderProvider.FromPortablePdbStream(stream, MetadataStreamOptions.LeaveOpen))
+            {
+                var pdbReader = pdbReaderProvider.GetMetadataReader();
+                var pdbSignature = new BlobContentId(pdbReader.DebugMetadataHeader.Id).Guid;
+                var pdbFileName = Path.GetFileName(pdbFullPath).ToLowerInvariant();
+                var pdbAge = "FFFFFFFF";
+                indexingPath = $"{pdbFileName}/{pdbSignature.ToString("N")}{pdbAge}/{pdbFileName}";
```
loic-sharma: >ToString [](start = 61, length = 8)

I believe this should be `ToUpperInvariant()`. See: [https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.Core/Services/SymbolIndexingService.cs#L161](https://github.com/loic-sharma/BaGet/blob/master/src/BaGet.Core/Services/SymbolIndexingService.cs#L161)

src/NuGet.Services.EndToEnd/Support/PackageType.cs
shishirx34: not much of a fan of ending in comma, but don't feel too strongly.. added.

src/NuGet.Services.EndToEnd/Support/PackageProperties.cs
shishirx34: removed.

src/NuGet.Services.EndToEnd/Support/PackageType.cs
joelverhagen: it is why this diff is very clear 😄

src/NuGet.Services.EndToEnd/Support/Utilities/PortableMetadataReader.cs
shishirx34: I don't think the casing matters here. I tried out all the casings and it worked fine when hitting the symbol server endpoint.

src/NuGet.Services.EndToEnd/Support/Package.cs
```diff
@@ -61,6 +72,30 @@ public static Package Create(PackageCreationContext context)
...
+                .Select(x => new PhysicalPackageFile(x.Stream) {
+                    TargetPath = $"lib/{x.FileName}"
+                });
+
+            ReadOnlyCollection<byte> nupkgBytes;
+            using (var nupkgStream = TestData.BuildPackageStreamForFiles(context, physicalPackageFilesList, properties.IsSymbolsPackage))
+            {
+                var bufferStream = new MemoryStream();
+                nupkgStream.CopyTo(bufferStream);
+                nupkgBytes = Array.AsReadOnly(bufferStream.ToArray());
```
joelverhagen: You don't need to do this. `ToArray()` copies the internal buffer.

src/NuGet.Services.EndToEnd/Support/Package.cs
joelverhagen: `CopyTo` here too

shishirx34: done.

src/NuGet.Services.EndToEnd/Support/Package.cs
shishirx34: caller needs readonly protection.

src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
shishirx34: done.

src/NuGet.Services.EndToEnd/SymbolsPackageTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            await _clients.SymbolServerClient.VerifySymbolFilesAvailable(
+                symbolsPackage.Id,
+                symbolsPackage.NormalizedVersion,
+                symbolsPackage.Properties.IndexedFiles,
+                _logger);
+        }
+
+        public void AssertPackageIsSymbolsPackage(Package package)
+        {
+            // The package should have indexed files if its a symbols package
```
loic-sharma: >its [](start = 56, length = 3)

it's

src/NuGet.Services.EndToEnd/Support/Package.cs
```diff
@@ -61,6 +72,30 @@ public static Package Create(PackageCreationContext context)
...
+            var physicalPackageFilesList = files
+                .Select(x => new { FileName = Path.GetFileName(x), Stream = GetMemoryStreamForFile(x) })
+                .Select(x => new PhysicalPackageFile(x.Stream) {
+                    TargetPath = $"lib/{x.FileName}"
+                });
+
+            ReadOnlyCollection<byte> nupkgBytes;
+            using (var nupkgStream = TestData.BuildPackageStreamForFiles(context, physicalPackageFilesList, properties.IsSymbolsPackage))
+            {
+                var bufferStream = new MemoryStream();
```
loic-sharma: >bufferStream [](start = 20, length = 12)

Considering that the `bufferStream` is copied into `nupkgBytes`, shouldn't you dispose this stream?

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -96,21 +110,21 @@ private async Task<Package> PreparePackagesAsync(PackageType requestedPackageTyp
...
                     }
                 }
 
                 await Task.WhenAll(pushTasks.Values);
-                
-                // Use the package that we just pushed.
+
+                // Use the packages that we just pushed.
                 return _packages[requestedPackageType];
             }
```
loic-sharma: Shouldn't this line be protected by the `_packagesLock`? Someone else could be mutating `_packages` here

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -210,14 +253,20 @@ private IEnumerable<PackageType> GetPackageTypes(PackageType requestedPackageTyp
                 .OrderBy(x => x);
         }
 
-        private PackageToPrepare InitializePackage(PackageType packageType)
+        /// <summary>
+        /// Return the list of <see cref="PackageToPrepare"/> in the ordered form
+        /// for which we want to run all the tasks synchronously
```
loic-sharma: >synchronously [](start = 51, length = 13)

`synchronously` -> `sequentially`

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
shishirx34: This is an existing code path, and I don't think it needs lock, the lock and the logic before ensures that the _packages is writable only once for the corresponding `requestedPackageType`.

loic-sharma: You cannot read a dictionary as it is being mutated, regardless of the key

shishirx34: The logic ensures that the dictionary mutation happens before any thread could possibly read it here. However, it doesn't seem to be protected by any mutation from outside of this method(we don't have any today). I'll add the lock for possible future degradation.

PR #79, opened 2019-03-15
Add "UseOfficialDns" to only determine search instance count from Azure management APIs

joelverhagen:
This means we don't use the direct cloud service (*.cloudapp.net) name when we are not running E2E tests for search service deployment. We use the official DNS found in the service index.

I ran into this problem because I deployed gallery to INT with the new SSL certificate before search. This should be fine.


joelverhagen:
With this setting, tests will hit `api-v2v3search-0.nuget.org:44301` instead of `nuget-prod-usnc-search.cloudapp.net:44301` which will avoid certain SSL issues.

PR #86, opened 2019-06-26
Add support for E2E tests for Azure search infrastructure.

shishirx34:
I have added minimal changes required for integration with the existing E2E tests. 

Once we get rid of the search cloud service, the E2E tests can be refactored to remove unnecessary configs. Meanwhile we need these to run the E2E tests on the older search service(cloud services).

I validated these changes with the new config in NuGetBuild Dev USNC and it the build passed just fine.

https://github.com/NuGet/Engineering/issues/2534

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/Clients/SearchServiceProperties.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/SearchServiceProperties.cs
@@ -6,19 +6,18 @@ using System;
 namespace NuGet.Services.EndToEnd.Support
 {
     /// <summary>
-    /// Represents a cloud search service with multiple instances.
+    /// Represents a cloud search service with multiple instances. Set the instance count to null to use the 
+    /// provided Uri as is, otherwise the port is appended based on the instance counts.
     /// </summary>
     public class SearchServiceProperties
     {
-        public SearchServiceProperties(Uri uri, int instanceCount, bool isAzureSearch = false)
+        public SearchServiceProperties(Uri uri, int? instanceCount)
         {
             Uri = uri;
             InstanceCount = instanceCount;
-            IsAzureSearch = isAzureSearch;
         }
 
         public Uri Uri { get; }
-        public int InstanceCount { get; }
-        public bool IsAzureSearch { get; }
+        public int? InstanceCount { get; }
     }
 }
--- a/src/NuGet.Services.EndToEnd/Support/Clients/SearchServiceProperties.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/SearchServiceProperties.cs
@@ -10,13 +10,15 @@ namespace NuGet.Services.EndToEnd.Support
     /// </summary>
     public class SearchServiceProperties
     {
-        public SearchServiceProperties(Uri uri, int instanceCount)
+        public SearchServiceProperties(Uri uri, int instanceCount, bool isAzureSearch = false)
         {
             Uri = uri;
             InstanceCount = instanceCount;
+            IsAzureSearch = isAzureSearch;
         }
 
         public Uri Uri { get; }
         public int InstanceCount { get; }
+        public bool IsAzureSearch { get; }
     }
 }
--- a/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
@@ -194,24 +194,36 @@ namespace NuGet.Services.EndToEnd.Support
             return searchServices;
         }
 
-        private async Task<SearchServiceProperties> GetSearchServiceFromAzureAsync(AzureCloudServiceDetails serviceDetails, ITestOutputHelper logger)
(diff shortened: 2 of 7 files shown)


joelverhagen:
Could you put a TODO in the code and track a GitHub item to clean up the cloud service stuff? A code comment would help in this case I think because the config object is kind of a frankenstein right now -- fine in the transition but not ideal for the long term.

src/NuGet.Services.EndToEnd/Support/Clients/SearchServiceProperties.cs
```diff
@@ -10,13 +10,15 @@ namespace NuGet.Services.EndToEnd.Support
     /// </summary>
     public class SearchServiceProperties
     {
-        public SearchServiceProperties(Uri uri, int instanceCount)
+        public SearchServiceProperties(Uri uri, int instanceCount, bool isAzureSearch = false)
```
joelverhagen: nit: no default so we can more easily see the cloud service codepaths too

src/NuGet.Services.EndToEnd/Support/Clients/SearchServiceProperties.cs
```diff
@@ -10,13 +10,15 @@ namespace NuGet.Services.EndToEnd.Support
...
         {
             Uri = uri;
             InstanceCount = instanceCount;
+            IsAzureSearch = isAzureSearch;
         }
 
         public Uri Uri { get; }
         public int InstanceCount { get; }
+        public bool IsAzureSearch { get; }
```
joelverhagen: `IsAzureSearch` seems like an implementation detail. Maybe we would make instance count nullable and just don't sent the port when instance count is null.

src/NuGet.Services.EndToEnd/Support/Configuration/GalleryConfiguration.cs
```diff
@@ -12,7 +12,7 @@ public class GalleryConfiguration
     {
         public string GalleryBaseUrl { get; set; }
 
-        public AzureCloudServiceDetails ServiceDetails { get; set; }
+        public AzureCloudServiceOrSearchDetails ServiceDetails { get; set; }
```
joelverhagen: Would it be possible to do this change without affecting the gallery config?

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -194,24 +207,39 @@ public async Task<IReadOnlyList<SearchServiceProperties>> GetSearchServicesAsync
...
-                                serviceDetails.ResourceGroup,
-                                serviceDetails.Name,
-                                serviceDetails.Slot,
-                                logger,
-                                CancellationToken.None);
-
-            var cloudService = AzureHelper.ParseCloudServiceProperties(result);
-
-            return new SearchServiceProperties(ClientHelper.ConvertToHttpsAndClean(cloudService.Uri), cloudService.InstanceCount);
+            if (serviceDetails.UseAzureSearchService)
```
joelverhagen: same as above, "AzureSearch" seems like an implementation detail. The distinction here seems to be "URLs in config" vs. "urls from Azure API"

src/NuGet.Services.EndToEnd/Support/Configuration/GalleryConfiguration.cs
shishirx34: This config will be used by the gallery e2e tests as well, so this is the minimum change required to modify it. Its an additive change so I don't think it should break anything existing. I have the PR for the new configs https://nuget.visualstudio.com/_git/NuGetBuild/pullrequest/978?_a=files&path=%2FE2EConfig%2FProd-TestGalleryUSNCStaging.json shows the gallery config

joelverhagen: that is on the `SearchServiceConfiguration` object though, not `GalleryConfiguration`

shishirx34: Oh you mean not rename the class?

joelverhagen: With the removed mentions of "azure search" and `UseConfiguredUrls`, maybe it's just `ServiceDetails` instead of `AzureCloudServiceDetails`. My resistance was having a type with the name "search" in it under `GalleryConfiguration`

shishirx34:
Verified the new changes with the E2E build and it looks good: https://nuget.visualstudio.com/NuGetBuild/_build/results?buildId=58249

src/NuGet.Services.EndToEnd/Support/Clients/SearchServiceProperties.cs
```diff
@@ -6,17 +6,18 @@
 namespace NuGet.Services.EndToEnd.Support
 {
     /// <summary>
-    /// Represents a cloud search service with multiple instances.
+    /// Represents a cloud search service with multiple instances. Set the instance count to null to use the
```
joelverhagen: `cloud or app service`

PR #87, opened 2019-07-05
Poll for some additional time against Azure Search Service

joelverhagen:
Address https://github.com/NuGet/NuGetGallery/issues/7320.

With a 30 second wait time introduced and no response caching, total run time went from **14 minutes and 34 seconds** to **38 minutes and 40 seconds**. With response caching the run time is **21 minutes and 29 seconds**.

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
@@ -25,6 +25,9 @@ namespace NuGet.Services.EndToEnd.Support
         private readonly TestSettings _testSettings;
         private readonly IRetryingAzureManagementAPIWrapper _azureManagementAPIWrapper;
 
+        private readonly SemaphoreSlim _pollCacheLock = new SemaphoreSlim(1);
+        private readonly Dictionary<string, V2SearchResponse> _pollCache = new Dictionary<string, V2SearchResponse>();
+
         public V2V3SearchClient(SimpleHttpClient httpClient, V3IndexClient v3IndexClient, TestSettings testSettings, IRetryingAzureManagementAPIWrapper azureManagementAPIWrapper)
         {
             _httpClient = httpClient;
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -323,37 +326,93 @@ private async Task<SearchServiceProperties> GetSearchServiceFromAzureAsync(Servi
...
+                    {
+                        logger.WriteLine($"Package {id} {version} is now complete. Polling for an additional {additionalPolling}.");
+                        sinceFirstComplete.Start();
+                    }
+                }
+                else
+                {
+                    if (sinceFirstComplete.IsRunning)
+                    {
+                        logger.WriteLine($"Package {id} {version} is no longer complete. Restarting the timer.");
```
loic-sharma: >Restarting [](start = 89, length = 10)

I'd suggest using the word `Resetting` instead of `Restarting`, considering that these operations mean different things on a C# stopwatch

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
```diff
@@ -323,37 +326,93 @@ private async Task<SearchServiceProperties> GetSearchServiceFromAzureAsync(Servi
             url = QueryHelpers.AddQueryString(url, "ignoreFilter", "true");
             url = QueryHelpers.AddQueryString(url, "semVerLevel", "2.0.0");
 
-            var duration = Stopwatch.StartNew();
+            // Determine if there is a cached response and if it is considered complete.
+            V2SearchResponse response;
             var complete = false;
-            do
+            var cached = false;
+            lock (_pollCacheLock)
```
loic-sharma: Consider using a semaphore for async lock

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
joelverhagen: Yeah, good point.

joelverhagen: Done.

src/NuGet.Services.EndToEnd/Support/Clients/V2V3SearchClient.cs
joelverhagen: Done.

PR #89, opened 2019-08-13
Early failure for missing configuration file.

agr:
Useful for running tests locally. Without it it fails with NPE later when accesses a configuration property.

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/TestSettings.cs
+++ b/src/NuGet.Services.EndToEnd/Support/TestSettings.cs
@@ -121,9 +121,17 @@ namespace NuGet.Services.EndToEnd.Support
 
         private static async Task<TestSettings> CreateInternalAsync(string configurationName)
         {
+            var configurationDirectoryName = Path.Combine(Environment.CurrentDirectory, "config");
+            var configurationFilename = configurationName + ".json";
+            var configurationFile = new FileInfo(Path.Combine(configurationDirectoryName, configurationFilename));
+            if (!configurationFile.Exists)
+            {
+                throw new InvalidOperationException($"Configuration file not found: {configurationFile.FullName}");
+            }
+
             var builder = new ConfigurationBuilder()
-             .SetBasePath(Path.Combine(Environment.CurrentDirectory, "config"))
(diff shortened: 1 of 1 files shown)


src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -121,9 +121,17 @@ private static async Task<TestSettings> CreateInternalAsync()
...
+            if (!configurationFile.Exists)
+            {
+                throw new InvalidOperationException($"Configuration file not found: {configurationFile.FullName}");
+            }
+
             var builder = new ConfigurationBuilder()
-             .SetBasePath(Path.Combine(Environment.CurrentDirectory, "config"))
-             .AddJsonFile(configurationName + ".json");
+                .SetBasePath(configurationDirectoryName)
+                .AddJsonFile(configurationFilename);
```
joelverhagen: this should have a "required" parameter that does the same thing, right?

scottbommarito: Yeah, it would be best if we didn't have to construct the path ourselves to verify if the file exists.

[There appears to be an `optional` parameter.](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.jsonconfigurationextensions.addjsonfile?view=aspnetcore-2.2#Microsoft_Extensions_Configuration_JsonConfigurationExtensions_AddJsonFile_Microsoft_Extensions_Configuration_IConfigurationBuilder_System_String_System_Boolean_) Have you tried that?

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -123,7 +123,7 @@ private static async Task<TestSettings> CreateInternalAsync(string configuration
         {
             var builder = new ConfigurationBuilder()
              .SetBasePath(Path.Combine(Environment.CurrentDirectory, "config"))
-             .AddJsonFile(configurationName + ".json");
+             .AddJsonFile(configurationName + ".json", optional: false);
```
shishirx34: nit: indentation.

PR #90, opened 2019-08-13
E2E tests for icons

agr:
Addresses https://github.com/NuGet/NuGetGallery/issues/7241

Two new package types were introduced:
* EmbeddedIconJpeg;
 * EmbeddedIconPng.

Since no client library that would be able to produce packages containing icon files is currently available, I had to resort to some minor hacks: `TestData.InjectIconMetadata` method was added to post-process the generated package and inject `<icon>` metadata into nuspec. Also, `AssemblyMetadataPackageFile` class was added to be able to add files to the package from assembly metadata.
`FlatContainerClient` was refactored a bit to allow binary file downloads as well as text with appropriate logging.


src/NuGet.Services.EndToEnd/IconTests.cs
```diff
@@ -0,0 +1,81 @@
...
+                    break;
+
+                case PackageType.EmbeddedIconPng:
+                    expectedExtension = ".png";
+                    break;
+
+                default:
+                    throw new ArgumentException($"Unsupported package type {packageType}");
+            }
+            var expectedFilename = $"icon{expectedExtension}";
```
joelverhagen: This is really "input extension" right? The URL in our endpoints has no file extension I thought

src/NuGet.Services.EndToEnd/IconTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            _pushedPackages = pushedPackages;
+            _clients = pushedPackages.Clients;
+            _logger = logger;
+        }
+
+        [Theory]
+        [InlineData(PackageType.EmbeddedIconJpeg)]
+        [InlineData(PackageType.EmbeddedIconPng)]
+        public async Task PushedPackageWithIconIsAvailableInV3(PackageType packageType)
```
joelverhagen: nit: can you separate the search/registration/flat container assertions into separate tests?

joelverhagen:
Could you put the word "jpg" and "png" in the associate images? Like in the actually picture? Will make future manual validations easier

src/NuGet.Services.EndToEnd/Support/AssemblyMetadataPackageFile.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Threading.Tasks;
+using NuGet.Packaging;
+using NuGet.Services.EndToEnd.Support.Utilities;
+
+namespace NuGet.Services.EndToEnd.Support
+{
+    /// <summary>
+    /// Provides files for <see cref="PackageBuilder"/> from test assembly metadata
+    /// </summary>
+    internal class AssemblyMetadataPackageFile : IPackageFile
```
joelverhagen: The name of this is kind of confusing. Why isn't this just "PackageFile"?

joelverhagen:
Can you do a E2E test run against Azure search before merging? 🐱

src/NuGet.Services.EndToEnd/Support/AssemblyMetadataPackageFile.cs
agr: There is a `PhysicalPackageFile` provided by the library indicating that it represents "physical" files (i.e. the ones that a present in a file system). `AssemblyMetadataPackageFile` represents "files" that come from the test assembly metadata.

agr:
@joelverhagen, https://nuget.visualstudio.com/NuGetBuild/_build/results?buildId=60633 - E2E test run against azure search.

PR #92, opened 2019-08-21
Poll all search-related endpoints (search+autocomplete) for availability

joelverhagen:
This is required while autocomplete is still going to Lucene but search is going to Azure Search. We need to poll both before continuing with the test since the "wait for search availability" method is used in both search and autocomplete tests. In other words, the tests assume that if a package is available in search it is available in autocomplete.

Theoretically we could have a different "poll for autocomplete" method but this has other complexities and in general the test's assumption is fine.

Required for https://github.com/nuget/nugetgallery/issues/7402.

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/Clients/V3IndexClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/V3IndexClient.cs
@@ -24,22 +24,34 @@ namespace NuGet.Services.EndToEnd.Support
             _testSettings = testSettings;
         }
 
-        public async Task<IReadOnlyList<string>> GetSearchBaseUrlsAsync(ITestOutputHelper logger)
+        public async Task<IReadOnlyList<string>> GetSearchGalleryQueryServiceUrlsAsync(ITestOutputHelper logger)
         {
             var v3Index = await GetV3IndexAsync(logger);
-            return GetResourceUrls(v3Index, t => t.Type.StartsWith("SearchGalleryQueryService/"));
+            return GetResourceUrls(v3Index, "SearchGalleryQueryService");
+        }
+
+        public async Task<IReadOnlyList<string>> GetSearchQueryServiceUrlsAsync(ITestOutputHelper logger)
+        {
+            var v3Index = await GetV3IndexAsync(logger);
+            return GetResourceUrls(v3Index, "SearchQueryService");
+        }
+
+        public async Task<IReadOnlyList<string>> GetSearchAutocompleteServiceUrlsAsync(ITestOutputHelper logger)
+        {
+            var v3Index = await GetV3IndexAsync(logger);
+            return GetResourceUrls(v3Index, "SearchAutocompleteService");
         }
 
         public async Task<IReadOnlyList<string>> GetFlatContainerBaseUrlsAsync(ITestOutputHelper logger)
         {
             var v3Index = await GetV3IndexAsync(logger);
-            return GetResourceUrls(v3Index, t => t.Type.StartsWith("PackageBaseAddress/"));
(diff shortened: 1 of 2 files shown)


src/NuGet.Services.EndToEnd/Support/Clients/V3IndexClient.cs
```diff
@@ -24,39 +24,57 @@ public V3IndexClient(SimpleHttpClient httpClient, TestSettings testSettings)
...
+                "RegistrationsBaseUrl/Versioned",
+                t => t.ClientVersion != null &&
                      NuGetVersion.Parse(t.ClientVersion) >= Version430Alpha);
         }
 
-        private static List<string> GetResourceUrls(V3Index v3Index, Func<Resource, bool> isMatch)
+        private static List<string> GetResourceUrls(V3Index v3Index, string resourceType, Func<Resource, bool> isMatch = null)
         {
+            isMatch = isMatch ?? (r => true);
+            Func<Resource, bool> isResourceTypeOrMatch = r =>
```
agr: Use local function here maybe?

src/NuGet.Services.EndToEnd/Support/Clients/V3IndexClient.cs
```diff
@@ -24,39 +24,57 @@ public V3IndexClient(SimpleHttpClient httpClient, TestSettings testSettings)
...
                 v3Index,
-                t => t.Type == "RegistrationsBaseUrl/Versioned" &&
-                     t.ClientVersion != null &&
+                "RegistrationsBaseUrl/Versioned",
+                t => t.ClientVersion != null &&
                      NuGetVersion.Parse(t.ClientVersion) >= Version430Alpha);
         }
 
-        private static List<string> GetResourceUrls(V3Index v3Index, Func<Resource, bool> isMatch)
+        private static List<string> GetResourceUrls(V3Index v3Index, string resourceType, Func<Resource, bool> isMatch = null)
```
agr: Should we just have separate overloads one for `.StartsWith($"{resourceType}/")` and another one for `isMatch`? This looks a cramming everything into the single method just to be able to call the same method in `GetSemVer2RegistrationBaseUrlsAsync`.

joelverhagen: Yeah, you're totally right. I'll do that.

src/NuGet.Services.EndToEnd/Support/Clients/V3IndexClient.cs
```diff
@@ -24,39 +24,57 @@ public V3IndexClient(SimpleHttpClient httpClient, TestSettings testSettings)
...
+            isMatch = isMatch ?? (r => true);
+            Func<Resource, bool> isResourceTypeOrMatch = r =>
+            {
+                return (r.Type == resourceType || r.Type.StartsWith($"{resourceType}/") && isMatch(r));
+            };
+
             return v3Index
                 .Resources
-                .Where(r => isMatch(r))
+                .Where(r => isResourceTypeOrMatch(r))
```
loic-sharma:
```suggestion
                .Where(isResourceTypeOrMatch)
```

src/NuGet.Services.EndToEnd/Support/Clients/V3IndexClient.cs
joelverhagen: Unnecessary with the other comment.

src/NuGet.Services.EndToEnd/Support/Clients/V3IndexClient.cs
joelverhagen: Done.

PR #97, opened 2019-09-12
Simplify base URL configuration for gallery and search

joelverhagen:
### Gallery URL configuration

What configuration is set and when?

- **always set**: `GalleryConfiguration.GalleryBaseUrl` - this is the official URL of gallery that is used for asserting things like license URL in V3 (which points to gallery).
- **set when testing gallery staging**: `GalleryConfiguration.ServiceDetails.BaseUrl` - this is the URL of gallery staging slot used during a gallery deployment.

When E2E test runs need to push packages or run other API calls against the gallery, they use the following precedence order:

1. `GalleryConfiguration.ServiceDetails.BaseUrl` - optional
1. `GalleryConfiguration.GalleryBaseUrl` - required

We never need cloud service details for gallery.

### Search URL configuration

What configuration is set and when?

- **set when deploying search staging**: `SearchServiceConfiguration.SingleSearchService` - this is the base URL for the staging slot used during a search service deployment. There are two variants of search.
  1. Azure Search - just the `BaseUrl` is set.
  1. Legacy Search - `BaseUrl` and cloud service details are set. The cloud service details are used to fetch the instance count. We need the instance count so we can poll all instances for package availability.
- **set when deploying non-search**: `SearchServiceConfiguration.IndexJsonMappedSearchServices` - this is a dictionary mapping all host names found in the service index to service details. There are two variants of the service details. In both cases the base URL used is the one found in the service index.
  1. Azure Search - no properties set.
  1. Legacy Search - Cloud service details are set to in order to fetch instance count.

Our code supports neither `SingleSearchService` or `IndexJsonMappedSearchServices` being set in which case we don't look up instances and use all of the search services from the service index. I don't know why we have this support... maybe for testing locally?

The following precedence order is used:

1. `IndexJsonMappedSearchServices`
1. `SingleSearchService`

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/Clients/IGalleryClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/IGalleryClient.cs
@@ -11,8 +11,20 @@ namespace NuGet.Services.EndToEnd.Support
 {
     public interface IGalleryClient
     {
-        Task<Uri> GetGalleryUrlAsync(ITestOutputHelper logger);
-        Uri GetGalleryBaseUrl(ITestOutputHelper logger);
+        /// <summary>
+        /// Get the gallery base URL used for interacting with the system. When deploying a component that is not
+        /// NuGetGallery, this will be the well-known public URL. When deploying NuGetGallery, this will be the staging
+        /// slot URL.
+        /// </summary>
+        Uri GetGalleryServiceBaseUrl();
+
+        /// <summary>
+        /// Get the well-known public URL of the NuGetGallery. This method should only be used when asserting the value
+        /// of a URL that is expected to point to the well known URL of NuGetGallery. Use the value returned by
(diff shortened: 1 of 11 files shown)


src/NuGet.Services.EndToEnd/Support/Clients/IGalleryClient.cs
```diff
@@ -11,8 +11,20 @@ namespace NuGet.Services.EndToEnd.Support
...
+        /// slot URL.
+        /// </summary>
+        Uri GetGalleryServiceBaseUrl();
+
+        /// <summary>
+        /// Get the well-known public URL of the NuGetGallery. This method should only be used when asserting the value
+        /// of a URL that is expected to point to the well known URL of NuGetGallery. Use the value returned by
+        /// <see cref="GetGalleryServiceBaseUrl"/> to interact with the gallery endpoints.
+        /// </summary>
+        Uri GetGalleryBaseUrl();
```
loic-sharma: Nice comments 👍

PR #98, opened 2019-09-12
Block TLS 1.0 and 1.1

scottbommarito:
https://github.com/NuGet/Engineering/issues/2617

E2E tests are defaulting to 1.0 and 1.1.


scottbommarito:
After restarting the build machines and installing .NET Framework 4.7.2, it is no longer defaulting to TLS 1.0 or 1.1

PR #100, opened 2019-10-15
Search test update to override expected registration roots

agr:
Progress on https://github.com/NuGet/Engineering/issues/2784

Currently, E2E tests [fail](https://nuget.visualstudio.com/NuGetBuild/_build/results?buildId=63169) when run against the non-default registration hive. This happens because one of the tests expects search service responses to contain registration URLs that are rooted to the URLs from service index files. Those are always going to be different when we migrate to a new registration hives.
This update introduces the configuration to allow optional override of expected registration roots.
Also a confusing variables were renamed in the affected test.

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/TestSettings.cs
+++ b/src/NuGet.Services.EndToEnd/Support/TestSettings.cs
@@ -126,7 +126,7 @@ namespace NuGet.Services.EndToEnd.Support
         /// <summary>
         /// Override base URLs for Semver 2 registration to be used in Azure Search tests.
         /// </summary>
-        public string[] SearchSemver2RegistrationRoots { get; set; }
+        public string[] SearchSemVer2RegistrationRoots { get; set; }
 
         private static async Task<TestSettings> CreateInternalAsync()
         {
--- a/src/NuGet.Services.EndToEnd/Support/TestSettings.cs
+++ b/src/NuGet.Services.EndToEnd/Support/TestSettings.cs
@@ -116,17 +116,17 @@ namespace NuGet.Services.EndToEnd.Support
         /// and search service configuration will not align at all times. This setting allows providing
         /// overrides for expected registration roots for Search Service tests.
         /// </remarks>
-        public string[] SearchBaseRegistrationRoots { get; set; }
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -106,6 +106,28 @@ public static async Task<TestSettings> CreateLocalTestConfigurationAsync(string
...
+        /// switching to (and testing) the alternative registration hives as service index information
+        /// and search service configuration will not align at all times. This setting allows providing
+        /// overrides for expected registration roots for Search Service tests.
+        /// </remarks>
+        public string[] SearchBaseRegistrationRoots { get; set; }
+
+        /// <summary>
+        /// Override base URLs for GZipped registration to be used in Azure Search tests.
+        /// </summary>
+        public string[] SearchGZipRegistrationRoots { get; set; }
```
scottbommarito: Is this used?

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
```diff
@@ -106,6 +106,28 @@ public static async Task<TestSettings> CreateLocalTestConfigurationAsync(string
...
+        /// Override base URLs for regular (Semver 1) registration to be used in Azure Search tests.
+        /// </summary>
+        /// <remarks>
+        /// Azure search service has registration root supplied through its own configuration
+        /// that it uses it responses (instead of taking one from the service index file). This makes
+        /// switching to (and testing) the alternative registration hives as service index information
+        /// and search service configuration will not align at all times. This setting allows providing
+        /// overrides for expected registration roots for Search Service tests.
+        /// </remarks>
+        public string[] SearchBaseRegistrationRoots { get; set; }
```
joelverhagen: Why not `BaseUrls` instead of `Roots`? This aligns with other properties in the config better.

Also `List<string>` would align more with the mutability properties on this config class.

src/NuGet.Services.EndToEnd/Support/TestSettings.cs
agr: No, I added it for completeness sake, but the test that is consuming those does not use that particular configuration bit.

PR #106, opened 2020-08-19
Use hijack=true to avoid search hijack in a more straight forward way

joelverhagen:
Progress on https://github.com/NuGet/Engineering/issues/2902.
This parameter was added here: https://github.com/NuGet/NuGetGallery/pull/8161

When we start blocking custom OData, we this pattern will break.

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
@@ -137,11 +137,11 @@ namespace NuGet.Services.EndToEnd.Support
         public async Task SearchPackageODataV2FromDBAsync(string id, string normalizedVersion, ITestOutputHelper logger)
         {
             var galleryEndpoint = GetGalleryServiceBaseUrl();
-            // Add 1 eq 1 to block the search hijack and execute the query against the database.
-            var urlRootAndPath = $"{galleryEndpoint}/api/v2/Packages()";
+            // Add the hijack=false query parameter to block the search hijack and execute the query against the database.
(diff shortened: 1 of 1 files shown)


src/NuGet.Services.EndToEnd/Support/Clients/GalleryClient.cs
```diff
@@ -137,11 +137,11 @@ public async Task DeprecateAsync(string id, IReadOnlyCollection<string> versions
         public async Task SearchPackageODataV2FromDBAsync(string id, string normalizedVersion, ITestOutputHelper logger)
         {
             var galleryEndpoint = GetGalleryServiceBaseUrl();
-            // Add 1 eq 1 to block the search hijack and execute the query against the database.
-            var urlRootAndPath = $"{galleryEndpoint}/api/v2/Packages()";
+            // Add the hijack=false query parameter to block the search hijack and execute the query against the database.
+            var urlRootAndPath = $"{galleryEndpoint}/api/v2/Packages(Id='{id}',Version='{normalizedVersion}')";
             var queryParameters = new Dictionary<string, string>()
             {
-                { "$filter", $"Id eq '{id}' and NormalizedVersion eq '{normalizedVersion}' and 1 eq 1" }
```
loic-sharma, line 144: Is this easily testable? If so, consider adding a test :)

joelverhagen: I think there is limited value in unit testing trivial parts of our end-to-end tests. They themselves are run as tests. If there is a complex test harness or some tricky code, I agree.

If you feel strongly, I'm happy to investigate it.

loic-sharma: Agreed, feel free to skip 👍

PR #116, opened 2022-05-18
Retry on Azure Search and CDN DNS resolution failure

joelverhagen:
Of the past 20 E2E test runs, this is the source of 2 of the failures.
It's not observed on our monitoring sources and always passes on a retry.
This is not worth blocking E2E tests on.

### DIFF
--- a/src/NuGet.Services.EndToEnd/Support/Clients/V3IndexClient.cs
+++ b/src/NuGet.Services.EndToEnd/Support/Clients/V3IndexClient.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 using NuGet.Versioning;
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.EndToEnd/Support/Clients/V3IndexClient.cs
```diff
@@ -104,10 +105,16 @@ private static List<string> GetResourceUrls(V3Index v3Index, Func<Resource, bool
...
+            return await RetryUtility.ExecuteWithRetry(
+                async () =>
+                {
+                    return await _httpClient.GetJsonAsync<V3Index>(
+                        _testSettings.V3IndexUrl,
+                        allowNotFound: false,
+                        logResponseBody: false,
+                        logger: logger);
+                },
+                ex => (ex.HasTypeOrInnerType<WebException>(out var we) && we.Status == WebExceptionStatus.NameResolutionFailure),
```
zhhyu, line 117: How often did it happen? Is it a concern?

zhhyu (approved):

src/NuGet.Services.EndToEnd/Support/Clients/V3IndexClient.cs
joelverhagen: Regarding, "Is it a concern?" generally speaking yes. Any E2E failure indicates a potential product issue. That being said, I looked at availability tests and CatchPoint tests and saw no similar symptom in the same period as the failures. Also, I don't recall any user reports about this problem, ever.

This suggests to me that it's related to our test agents and not a real customer impacting issue. IMO, we can depend on monitoring for this sort of failure and this change will simply help our deployment agility.

PR #119, opened 2022-07-08
Update package version

lyndaidaii:
Update newtonsoft.json/Servercommon/NuGet.Client SDK

Address: [#4452](https://github.com/NuGet/Engineering/issues/4452)

### DIFF
--- a/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
+++ b/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
@@ -157,11 +157,14 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder">
       <Version>2.2.0</Version>
     </PackageReference>
+    <PackageReference Include="Newtonsoft.Json">
+      <Version>13.0.1</Version>
+    </PackageReference>
     <PackageReference Include="NuGet.Protocol">
-      <Version>4.9.2</Version>
+      <Version>6.2.1</Version>
     </PackageReference>
     <PackageReference Include="NuGet.Services.Configuration">
-      <Version>2.72.0</Version>
+      <Version>2.103.0</Version>
     </PackageReference>
     <PackageReference Include="System.Reflection.Metadata">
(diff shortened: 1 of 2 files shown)


src/NuGet.Services.EndToEnd/Support/AssemblyMetadataPackageFile.cs
```diff
@@ -28,6 +29,8 @@ public AssemblyMetadataPackageFile(string resourceName)
 
         public FrameworkName TargetFramework { get; set; }
 
+        public NuGetFramework NuGetFramework { get; set; }
```
shishirx34, line 32: why is this needed?

src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
```diff
@@ -157,11 +157,14 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder">
       <Version>2.2.0</Version>
     </PackageReference>
+    <PackageReference Include="Newtonsoft.Json">
+      <Version>13.0.1</Version>
+    </PackageReference>
```
shishirx34, line 162: since you are adding a top level dependency, doesn't this need binding redirect for transitive dependencies of Newtosoft.Json?

agr: I think we only need explicit binding redirects only for web projects (that use `web.config` instead of `app.config`).

joelverhagen: That's correct.

src/NuGet.Services.EndToEnd/Support/AssemblyMetadataPackageFile.cs
joelverhagen: The `IPackageFile` interface changed via the NuGet client package update.

joelverhagen (approved):

PR #131, opened 2025-04-22
Add unicode package Id testing

erdembayar:
Address: https://github.com/NuGet/Engineering/issues/5850

### DIFF
--- a/src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
+++ b/src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
@@ -24,8 +24,11 @@ namespace NuGet.Services.EndToEnd
 
         [GalleryTestTheory]
         [InlineData(PackageType.SemVer1Stable, true, null)]
+        [InlineData(PackageType.SemVer1StableUnicodeId, true, null)]
         [InlineData(PackageType.SemVer1Stable, false, null)]
+        [InlineData(PackageType.SemVer1StableUnicodeId, false, null)]
         [InlineData(PackageType.SemVer2Prerel, true, "2.0.0")]
+        [InlineData(PackageType.SemVer2PrerelUnicodeId, true, "2.0.0")]
         [InlineData(PackageType.SemVer2StableMetadata, false, "2.0.0")]
         public async Task V2PackageVersionsAutocompleteResultsMatchV3Results(
             PackageType packageType,
--- a/src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
+++ b/src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
@@ -26,9 +26,7 @@ namespace NuGet.Services.EndToEnd
         [InlineData(PackageType.SemVer1Stable, true, null)]
         [InlineData(PackageType.SemVer1StableUnicodeId, true, null)]
         [InlineData(PackageType.SemVer1Stable, false, null)]
-        [InlineData(PackageType.SemVer1StableUnicodeId, false, null)]
         [InlineData(PackageType.SemVer2Prerel, true, "2.0.0")]
-        [InlineData(PackageType.SemVer2PrerelUnicodeId, true, "2.0.0")]
         [InlineData(PackageType.SemVer2StableMetadata, false, "2.0.0")]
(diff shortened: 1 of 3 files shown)


src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
```diff
@@ -103,5 +103,44 @@ public async Task V2PackageIdsAutocompleteResultsMatchV3Results(
                 Assert.Equal(v2Response, v3Response.Data);
             }
         }
+
+        [GalleryTestTheory]
+        [InlineData(PackageType.SemVer1Stable, true, null)]
```
joelverhagen, line 108: I think you'd want a new package type here, like `PackageType.UnicodeId` or something.

src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
```diff
@@ -103,5 +103,44 @@ public async Task V2PackageIdsAutocompleteResultsMatchV3Results(
...
+        [InlineData(PackageType.SemVer1Stable, false, null)]
+        public async Task UnicodePackageIdV2PackageVersionsAutocompleteResultsMatchV3Results(
+            PackageType packageType,
+            bool includePrerelease,
+            string semVerLevel)
+        {
+            // Arrange
+            var package = await _pushedPackages.PrepareAsync(packageType, _logger, unicodeId : true);
```
joelverhagen, line 116: I don't see where this new parameter is used.

erdembayar: Sorry, I didn't stage one more file change.

src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
erdembayar: I pushed another file

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -281,7 +281,7 @@ private async Task<List<PackageToPrepare>> InitializePackagesAsync(PackageType p
                 case PackageType.SemVer2DueToSemVer2Dep:
                     packageToPrepare = new PackageToPrepare(Package.Create(new PackageCreationContext
                     {
-                        Id = id,
+                        Id = unicodeId ? id + "пакет包elsökning" : id,
```
joelverhagen, line 284: This is the wrong approach. The `PackageType.<Type>` is reused for other tests. We should introduce a new package type and test on that, instead of modifying the existing package type.

erdembayar: Sure, added new type.

src/NuGet.Services.EndToEnd/Support/PushedPackagesFixture.cs
```diff
@@ -319,6 +319,11 @@ private async Task<List<PackageToPrepare>> InitializePackagesAsync(PackageType p
                     packageToPrepare = new PackageToPrepare(Package.Create(packageType, id, SemVer2PrerelVersion));
                     break;
 
+                case PackageType.SemVer2PrerelUnicodeId:
+                    // package Id contains unicode characters.
+                    packageToPrepare = new PackageToPrepare(Package.Create(packageType, id + "пакет包elsökning123", SemVer2PrerelVersion));
```
erdembayar, line 324: I don't believe passing actual package Id from test adds any value, make code complicated, this this should be suffice. We have other unit test with more cases in NuGetJobs solution.

src/NuGet.Services.EndToEnd/AutoCompleteResultTests.cs
```diff
@@ -24,8 +24,11 @@ public AutocompleteResultTests(PushedPackagesFixture pushedPackages, ITestOutput
 
         [GalleryTestTheory]
         [InlineData(PackageType.SemVer1Stable, true, null)]
+        [InlineData(PackageType.SemVer1StableUnicodeId, true, null)]
```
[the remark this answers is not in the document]
erdembayar: Sure, removed other 2

mariaghiondea (approved):

PR #132, opened 2025-08-11
Update Azure.Identity to v1.11.4

etvorun:
Fixes https://github.com/NuGet/Engineering/issues/5975

### DIFF
--- a/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
+++ b/src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
@@ -152,7 +152,6 @@
     </None>
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


joelverhagen (dismissed):

src/NuGet.Services.EndToEnd/NuGet.Services.EndToEnd.csproj
```diff
@@ -152,6 +152,7 @@
     </None>
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Azure.Identity" Version="1.11.4"/>
```
joelverhagen, line 155: Ideally we would update this dependency without lifting it (perhaps by updating the direct dependency that pulls it in). But if that's not feasible, this solution is fine.

